PR #18, opened 2015-01-20
Automatically transition Solutions tickets to open status for OpenCraft

antoviaque:
@singingwolfboy Does that look right?

### DIFF
--- a/openedx_webhooks/views/jira.py
+++ b/openedx_webhooks/views/jira.py
@@ -149,6 +149,7 @@ def should_transition(issue):
         "edx-employees": set(("ALL",)),
         "clarice": set(("MOB",)),
         "bnotions": set(("MOB",)),
+        "opencraft": set(("SOL",)),
     }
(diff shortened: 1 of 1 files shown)


edx-webhook:
Thanks for the pull request, @antoviaque! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request. 

To automatically create an OSPR issue for this pull request, just visit this link: http://openedx-webhooks.herokuapp.com/github/process_pr?number=18&repo=edx%2Fopenedx-webhooks

singingwolfboy:
Looks great! Thanks!

nedbat:
We should really get this information into an org.yaml file.

PR #19, opened 2015-02-06
We aren't interested in synchronize events, ignore them.

nedbat:


nedbat:
@singingwolfboy I think it's better to be OK with these events, rather than sending errors back to GitHub.

PR #20, opened 2015-02-13
Remove CacheControl

singingwolfboy:
It's just making the bot less reliable. @sarina, can you review this?


nedbat:
How does CacheControl make the bot less reliable?

sarina:
@nedbat - it doesn't recognize new labels in the repo, and a few other things that I can't remember, but have always been solvable by restarting the bot so it's forced to clear its cache.

:+1:

sarina:
(if this causes other issues we can always revert it)

nedbat:
:+1:

PR #22, opened 2015-03-09
Simplify logic for determining transition state

sarina:
@singingwolfboy conditional loops omg

### DIFF
--- a/openedx_webhooks/views/jira.py
+++ b/openedx_webhooks/views/jira.py
@@ -188,18 +188,20 @@ def issue_opened(issue, bugsnag_context=None):
         # goes to, so a transition to go from "Open" to "In Progress" may be
         # named something like "Start Work".
         transitions = {t["name"]: t["id"] for t in transitions_resp.json()["transitions"]}
-        # We attempt to transition the issue into the "Open" state, so we
-        # look for a transition named "Open"
-        if "Open" in transitions:
-            new_status = "Open"
-        # UX uses "Design Backlog" instead of "Open", so check for that instead
-        elif "Design Backlog" in transitions:
-            new_status = "Design Backlog"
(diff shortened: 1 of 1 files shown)


openedx_webhooks/views/jira.py
```diff
@@ -188,18 +188,17 @@ def issue_opened(issue, bugsnag_context=None):
...
-            new_status = "Design Backlog"
-        # If it's an OSPR subtask (used by teams to manage reviews), transition to team backlog
-        elif to_unicode(issue["fields"]["project"]["key"]) == "OSPR" and issue["fields"]["issuetype"]["subtask"]:
-            new_status = "To Backlog"
+        # We attempt to transition the issue into the "Open" state for the given project
+        # (some projects use a different name), so look for a transition with the right name
+        for state_name in ["Open", "Design Backlog", "To Do"]:
+            if state_name in transitions:
+                new_status = state_name
         else:
```
singingwolfboy: You've got a `for`/`else` block here, which I don't think will do what you want it to do. Python will only execute the `else` block if the `for` block never executes -- and it will *always* execute, based on how you've written this code. I think you wan to do something more like:
```python
new_status = None
from transition_name in ("Open, "Design Backlog", "To Do"):
    if transition_name in transitions:
        new_status = state_name
is_ospr = to_unicode(issue["fields"]["project"]["key"]) == "OSPR"
is_subtask = issue["fields"]["issuetype"]["subtask"]
if not new_status and is_ospr and is_subtask:
    new_status = "To Backlog"

if not new_status:
    raise ValueError("No valid transition! Possibilities are {}".format(transitions.keys()))
```

sarina: Word - yes. I did a bit differently so that we don't check the is_ospr/is_subtask logic if we already have a new status.

singingwolfboy:
:+1:

PR #24, opened 2015-05-20
Transition items to 'In Backlog' instead of 'Open'

benpatterson:
We should transition items that are no longer in triage to simply 'In Backlog'. Setting it to open implies that it's getting work done on it, which is not the default case.

---end commit message---

This would work better for me. CCing some folks as they may have opinions and would likely be affected by this change.

CC @jzoldak @singingwolfboy @dsjen @cahrens @feanil @cpennington [and zach...don't know his github username]

### DIFF
--- a/openedx_webhooks/views/jira.py
+++ b/openedx_webhooks/views/jira.py
@@ -193,7 +193,7 @@ def issue_opened(issue, bugsnag_context=None):
         # (some projects use a different name), so look for a transition with the right name
         new_status = None
         action = None
-        for state_name in ["Open", "Design Backlog", "To Do"]:
+        for state_name in ["In Backlog", "Design Backlog", "To Do"]:
             if state_name in transitions:
                 new_status = state_name
                 action = "Transitioned to '{}'".format(state_name)
@@ -221,7 +221,7 @@ def issue_opened(issue, bugsnag_context=None):
 
     # log to stderr
     if transitioned and not action:
-        action = "Transitioned to Open"
+        action = "Transitioned to Backlog"
     else:
         action = "ignored"
     print(


rock345:
:+1:

dsjen:
Love it! :+1:

cahrens:
:-1: "In Backlog" is supposed to mean that an item is groomed, has story points assigned, and is ready to be picked up by the team. I don't think Open means that work is getting done on it at all-- it's the state before it goes into "Ready For Grooming", before anyone on the team has examined if it is really something we want to do.

dsjen:
Do we currently have an "in backlog" state or is that what the ideal state of "in backlog" should be?

singingwolfboy:
> it's the state before it goes into "Ready For Grooming", before anyone on the team has examined if it is really something we want to do.

@cahrens: you know, that sounds to me like it should stay in "Needs Triage". Because when your team is determining if it's something that you want to do, that is triaging the issue.

cahrens:
@singingwolfboy I thought "Needs Triage" was only used by the Open Source team to look at tickets being created by the open source community, so they didn't flood our backlogs. I am feeling no pain with the current states and don't see a reason to change them.

And yes, we do currently have an "In Backlog" state. That's the state a ticket goes into when you drag it into the next column (to the right) from "In Grooming" on the product discovery backlog. You should do this only after points have been assigned. The vast majority of all your tickets on your team backlog will be in the state "In Backlog".

singingwolfboy:
@cahrens: you're right, the "Needs Triage" state was originally conceived of as a way to filter tickets created by the open source community, so that the open source team could make sure that they contained appropriate information and were filed in the appropriate place before other teams look at them. It's simply that the way you described what your team was doing, sounded a bit like what my team was doing for those tickets. If it simpler not to change anything about the "Needs Triage" state, then let's not change it.

I should also point out that every team organizes their JIRA issues a bit differently, and one team's definition of "Open" may be different from another team's definition of "Open". @benpatterson, would it be easier to have the bot move issues to different states depending on the project? For example, it could move T&L tickets to "Open", but move TestEng tickets to "In Backlog". The downside of that approach is that it makes the logic in this project more confusing, and perhaps it would be better to standardize on these definitions, instead.

cahrens:
I think Miki's desire is to standardize the process across teams.

benpatterson:
I'd like to change this somehow. The current system doesn't work for me, and I've seen other teams go through the same pain. Once the ticket is created, it's immediately set to open, but open typically represents things that are being actively worked on.

Current workflow for myself and some other teams is: create the ticket, then go back into the ticket you just created, and change its state to be more accurate. That's overhead I'd love to get rid of!

I could see going to the 'ready for grooming' state, though our team would need to make changes in order for that to be effective. We don't currently use that state, but I can see the usefulness of it. Except of course, when we are actually grooming...then it would go straight to backlog.

cahrens:
I don't understand why "Open" represents the state of things being actively worked on. Those should be "In Progress".

benpatterson:
Ah that's a good point, @cahrens. I guess in my mind, if it's not in some explicit backlog state, then something's being done with it.

benpatterson:
I sent an e-mail offline so we can work this out. Hat-tip to @singingwolfboy

benpatterson:
I'll leave this open while it gets sorted out offline.

PR #26, opened 2015-07-30
responses have a raise_for_status() method that is neat

singingwolfboy:
http://docs.python-requests.org/en/latest/api/?highlight=raise_for_status#requests.Response.raise_for_status


nedbat:
:+1:

sarina:
Don't know how to test but looks fine :+1:

PR #31, opened 2015-08-11
Werkzeug ProxyFix is awesome

singingwolfboy:


singingwolfboy:
@sarina @nedbat this change allows Flask to determine when it's served at an HTTPS endpoint, even when that endpoint is behind a proxy. I just discovered this ProxyFix thing a few days ago. Can one or both of you review this?

sarina:
:+1: looks cool

PR #32, opened 2015-08-11
Chunk the rescan repos job so it doesn't time out.

sarina:
@singingwolfboy can you take a quick look please.


singingwolfboy:
Eww, ew ew ew. But I understand why we're doing this. :disappointed: :+1:

sarina:
Yes I think we'll have to do some significant code refactoring to add a worker, and part of that refactor can include removing this hack.

PR #34, opened 2015-08-27
Pull from private sekkrit repo

sarina:
@nedbat


sarina:
OK I'm going to merge this and see if it causes any issues.

PR #36, opened 2015-08-27
Pull from private sekkrit repo

sarina:
@nedbat @singingwolfboy 

We need to do this; how can the bot read from a private repo?

I tried merging as is but it failed:

screen shot 2015-08-27 at 4 31 04 pm

https://bugsnag.com/openedx-webhooks/openedx-webhooks-1/errors/55df72d98c759b9d281a0c37?event_id=55df72d97a636bc5291cb898


sarina:
@singingwolfboy ping on this

singingwolfboy:
@sarina that error message looks like a permissions error from GitHub. The GitHub user that the bot runs under does not have permission to view that repository. It has nothing to do with the bot's code.

sarina:
@singingwolfboy yes, I understand that. My question as stated in the PR description is: "We need to do this; how can the bot read from a private repo?"

singingwolfboy:
@sarina it looks like the `edx-webhook-admin` team is the one we've been using for this purpose. You can add repos to that team, and the users in that team will get admin access to those repos. The only user in that team is @edx-webhook, so that should be what you need.

singingwolfboy:
It would be nice if we had an organization-level webhook such that any time a repo is added to the edX organization on GitHub, it is also added to this team. GitHub provides organization-level webhooks, so we could totally do this, we just haven't yet.

sarina:
Yeah that would be a good idea. OK I'm going to try to merge this now and see what happens.

sarina:
thanks that worked!

PR #37, opened 2015-08-31
Major refactor for Celery task queue support

singingwolfboy:
* Introduce celery dependency
* Pull github views, jira views, and UI into separate blueprints
* Application factory pattern
* Added .env.example file, for ease of running `foreman start` locally

This PR does not cause openedx-webhooks to actually run jobs on the Celery backend yet, it simply sets up the framework to do so. The Celery dependency means that we'll need to install the RabbitMQ Bigwig addon on Heroku.

@nedbat @sarina review?


.env.example
```diff
@@ -0,0 +1,6 @@
+OPENEDX_WEBHOOKS_CONFIG=development
+OAUTHLIB_INSECURE_TRANSPORT=1
+GITHUB_OAUTH_CLIENT_ID=
+GITHUB_OAUTH_CLIENT_SECRET=
+JIRA_OAUTH_CONSUMER_KEY=
+JIRA_OAUTH_RSA_KEY=
```
sarina: what is this file

sarina: ie maybe add a preamble to explain

sarina:
how are you testing this?

.env.example
singingwolfboy: An example environment file for Foreman to use. The idea is that a developer can copy `.env.example` to `.env`, put in real tokens (which are sensitive information, and should *not* be committed to the GitHub repository), and then run `foreman start` to run both the web process and the worker process. See also: http://ddollar.github.io/foreman/#ENVIRONMENT

singingwolfboy:
@sarina no automated tests yet. I tested that I was able to run the project locally using `foreman start`, and that the UI pages successfully render.

.env.example
singingwolfboy: I added a preamble

sarina: ty

sarina:
I don't see anything that jumps out as problematic here, but don't necessarily feel like I can give a full thumbs-up. What is your plan?

singingwolfboy:
@sarina roll it out and monitor it, same as we've done in the past. Some day we'll decide that we need to focus on some real test coverage for this project, it's a question of when the benefit of doing so outweighs the cost of doing all the annoying mocking that will be required. Regardless, when that happens, it should happen in a separate pull request. If you think that this project needs better tests before we can merge this, then I can work on that, but it will take longer for us to get our pull request bot back into a working state, which I don't think is the right trade-off here.

sarina:
I agree, just want to know what your game plan is. I'm fine with that. When
would you plan to merge and test? I would suggest earlier in the day is
best.

singingwolfboy:
I'll do it tomorrow morning.

PR #38, opened 2015-08-31
Build bot comments using Jinja2 templates

singingwolfboy:
@sarina

### DIFF
--- a/tests/test_pr_comments.py
+++ b/tests/test_pr_comments.py
@@ -1,0 +1,51 @@
+import pytest
+from datetime import datetime
+import openedx_webhooks
+from openedx_webhooks.views.github import (
+    github_community_pr_comment, github_contractor_pr_comment
+)
+
+#pytest.mark.usefixtures('betamax_session')
+
+def make_pull_request(user, number=1, base_repo_name="edx/edx-platform", head_repo_name="edx/edx-platform", created_at=None):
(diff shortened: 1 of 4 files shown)


tests/test_pr_comments.py
```diff
@@ -0,0 +1,51 @@
+import pytest
+from datetime import datetime
+import openedx_webhooks
+from openedx_webhooks.views.github import (
+    github_community_pr_comment, github_contractor_pr_comment
+)
+
+#pytest.mark.usefixtures('betamax_session')
```
sarina: ?

tests/test_pr_comments.py
```diff
@@ -0,0 +1,51 @@
...
+
+def make_pull_request(user, number=1, base_repo_name="edx/edx-platform", head_repo_name="edx/edx-platform", created_at=None):
+    "this should really use a framework like factory_boy"
```
sarina: "this is not a real docstring"

sarina:
:+1:

singingwolfboy:
I've discovered that this doesn't work the way I thought it would. The linebreaks in the Jinja2 templates are preserved, and cause the GitHub comments to get formatted weirdly. I might need to write a custom Jinja2 extension to indicate that linebreaks should be stripped -- that sounds like an interesting side project. I'm going to close this PR for now, but keep the branch around.

PR #39, opened 2015-09-01
Jinja2 template for comments, and fix tests for private repo

singingwolfboy:
Re-doing #38 -- I discovered that I can use `{% filter replace("\n", " ") %}` to get rid of the newlines in the Jinja2 templates. Also, Betamax needed some additional logic to be able to fetch the fixtures for the tests.


singingwolfboy:
@sarina can you review this? It's very similar to #38.

sarina:
:+1:

nedbat:
Why do .md files need the newlines removed?

singingwolfboy:
@nedbat because GitHub renders newlines as `<br>` HTML elements, it seems.

PR #40, opened 2015-09-01
Use Sentry instead of Bugsnag

singingwolfboy:
@nedbat @sarina review?

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -11,6 +11,7 @@ Install Add-ons
 This project uses:
 
 * `Heroku Postgres <https://addons.heroku.com/heroku-postgresql>`_
+* `RabbitMQ Bigwig <https://addons.heroku.com/rabbitmq-bigwig>`_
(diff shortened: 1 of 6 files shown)


sarina:
docs/install.rst:14:* `Bugsnag <https://addons.heroku.com/bugsnag>`_

sarina:
I'm OK with this change, but I'd like to hold off on it until we're sure the other changes are working. I'm anxious about merging a complete change to our bug manager at the exact same time we've made major, potentially breaking, changes.

singingwolfboy:
@sarina good catch. I've updated the docs.

I understand the concern about changing the bug manager right now, but the reason I'm doing it is because it simplifies the function signatures (no more passing around a `bugsnag_context` dictionary), which will be important for taking these functions and running them as worker processes. The simpler we can make this code, the less likely it is that anything will go wrong.

sarina:
Hmm well I'm ok if Ned's ok.

PR #41, opened 2015-09-01
Pass auth to is_internal_pull_request

singingwolfboy:
This should resolve the Sentry error we saw. @sarina @nedbat


sarina:
:+1:

PR #43, opened 2015-09-02
Move GitHub tasks to Celery worker

singingwolfboy:
Move GitHub tasks to a Celery worker

### DIFF
--- a/openedx_webhooks/github_views.py
+++ b/openedx_webhooks/github_views.py
@@ -18,7 +18,7 @@ from openedx_webhooks.info import (
     get_people_file, get_repos_file,
     is_internal_pull_request, is_contractor_pull_request,
     )
-from openedx_webhooks.utils import paginated_get
+from openedx_webhooks.utils import paginated_get, minimal_wsgi_environ
 from openedx_webhooks.tasks.github import (
     pull_request_opened, pull_request_closed, rescan_repository
 )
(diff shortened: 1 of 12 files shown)


singingwolfboy:
OK, I've verified that this branch runs successfully, and I can rescan repos successfully. @sarina @nedbat can you two review it now?

openedx_webhooks/github_views.py
```diff
@@ -48,14 +47,24 @@ def pull_request():
...
         return "Ignoring synchronize events from github", 200
+    if event["action"] == "opened":
+        result = pull_request_opened.delay(pr, wsgi_environ=minimal_wsgi_environ())
+        status_url = url_for("tasks.status", task_id=result.id, _external=True)
+        resp = jsonify({"message": "queued", "status_url": status_url})
+        resp.status_code = 202
+        resp.headers["Location"] = status_url
+        return resp
+    if event["action"] == "closed":
+        result = pull_request_closed.delay(pr)
```
sarina: This is the only line that seems to differ from the code under `if event["action"] == "opened"` so maybe,

```python
if event["action"] == "opened":
    result = pull_request_opened.delay(pr, wsgi_environ=minimal_wsgi_environ())
elif event["action"] == "closed":
    result = pull_request_closed.delay(pr)
else:
    # some error message? unknown action?
status_url = url_for("tasks.status", task_id=result.id, _external=True)
....
return resp
```

openedx_webhooks/tasks/github.py
```diff
@@ -0,0 +1,336 @@
...
+        logger.info(msg)
+        return msg
+
+    repo = pr["base"]["repo"]["full_name"].decode('utf-8')
+    people = get_people_file(session=github)
+    custom_fields = get_jira_custom_fields(jira)
+
+    if user in people:
+        user_name = people[user].get("name", "")
+    else:
```
sarina: Perhaps better as:
```python
user_name = people[user].get("name", "")
if not user_name:
    user_resp = github.get(pr["user"]["url"])
    ...
```

this way we don't end up with a blank user_name ever, at worst we get their github username.

sarina:
Seems pretty reasonable to me

singingwolfboy:
@sarina I responded to all your feedback -- thank you! @nedbat, do you have any suggestions?

PR #44, opened 2015-09-02
Stop duplicate contractor comments

singingwolfboy:
@sarina

### DIFF
--- a/openedx_webhooks/github_views.py
+++ b/openedx_webhooks/github_views.py
@@ -189,8 +189,9 @@ def install():
 
 
 @memoize
-def github_whoami():
-    self_resp = github.get("/user")
+def github_whoami(session=None):
+    session = session or github
+    self_resp = session.get("/user")
     rate_limit_info = {k: v for k, v in self_resp.headers.items() if "ratelimit" in k}
     print("Rate limits: {}".format(rate_limit_info), file=sys.stderr)
     self_resp.raise_for_status()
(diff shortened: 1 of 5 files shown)


openedx_webhooks/github_views.py
```diff
@@ -455,6 +460,23 @@ def github_contractor_pr_comment(pull_request):
...
+    my_username = me["login"]
+    comment_url = "/repos/{repo}/issues/{num}/comments".format(
+        repo=pull_request["base"]["repo"]["full_name"].decode('utf-8'),
+        num=pull_request["number"],
+    )
+    for comment in paginated_get(comment_url, session=github):
+        # I only care about comments I made
+        if comment["user"]["login"] != my_username:
+            continue
+        magic_phrase = "It looks like you're a member of a company that does contract work for edX."
```
nedbat: Is there any way to share this text with the code that makes the comment, so they won't get out of sync?

openedx_webhooks/github_views.py
```diff
@@ -455,6 +460,23 @@ def github_contractor_pr_comment(pull_request):
     )
 
 
+def has_contractor_comment(pull_request, session=None):
```
nedbat: As this code gets more complex, it would be helpful to write docstrings.

singingwolfboy: Added.

openedx_webhooks/github_views.py
singingwolfboy: I don't think there's a way to share the text, unfortunately. But we can write tests!

sarina: Yeah I'm concerned about this

sarina: Alternatively, a simple test that tests that `magic_phrase` is in the contractor comment, so if we change the contractor comment the test would flag this?

singingwolfboy: I added a test that asserts that after the bot has made a comment, the `has_contractor_comment` function returns True.

sarina:
:+1: once tests pass

PR #45, opened 2015-09-04
Move JIRA rescan user task to task queue

singingwolfboy:
Our load on the webservers is still very spiky, and it appears to be spiking once per hour. I'm betting it's because of the "rescan JIRA users" task which is run once every hour. Moving this to the task queue will smooth our the load on the webservers, which will hopefully make them more reliable.


singingwolfboy:
I verified that this works correctly on `openedx-webhooks-test.herokuapp.com`. (I just had to set the correct enviornment variables first.)

PR #46, opened 2015-09-25
Use the name of the end state rather than the name of the transition

singingwolfboy:
In JIRA, you can name a state transition whatever you want -- the default name
is the name of the end state, but you can change that. It looks like the API
does expose the name of the end state, so we should key off of that, in case
the transition has a different name from the end state.

Note that this assumes that there will never be more than one transition ending
at the same state. If that happens, one of them will be clobbered. (It's not
clear that we have a good way of dealing with that even before this change.)

### DIFF
--- a/openedx_webhooks/jira_views.py
+++ b/openedx_webhooks/jira_views.py
@@ -183,7 +183,7 @@ def issue_opened(issue):
         # Note that a transition may not have the same name as the state that it
         # goes to, so a transition to go from "Open" to "In Progress" may be
         # named something like "Start Work".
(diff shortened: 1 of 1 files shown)


sarina:
I'm reverting this. Too many team's workflows would need to be changed and renamed to make this work. I changed the Open workflow, but the Accessibility flow is also broken and I don't want to have to be triaging every new random project that has been working but is not.

It's also easier and less disruptive for me to rename (or add in!) a named transition for tasks to follow; changing a state's name is wildly disruptive, and requires also editing multiple board's column views, which is just crazy.

PR #48, opened 2015-10-16
Ignore external OSPR issues

singingwolfboy:
@sarina: This should make the bot stop complaining about external pull requests. However, it relies on JIRA issues having the "repo" field filled out correctly. If the field is empty, or filled out incorrectly, then there's not really anything the bot can do but throw an error. I've corrected the "repo" field on [OSPR-850](https://openedx.atlassian.net/browse/OSPR-850).


sarina:
sounds good, loooks good, ty :+1:

PR #51, opened 2015-10-19
get_jira_issue() returns a requests.Response object

singingwolfboy:
should resolve https://app.getsentry.com/edx-1/openedx-webhooks/group/88968953/


singingwolfboy:
Closed in favor of #55

PR #53, opened 2015-10-28
Revert "Use the name of the end state rather than the name of the tra…

sarina:
Reverts edx/openedx-webhooks#46

This is breaking too many things and was a poorly conceived "improvement"


sarina:
I'm reverting this. Too many team's workflows would need to be changed and renamed to make this work. I changed the Open workflow, but the Accessibility flow is also broken and I don't want to have to be triaging every new random project that has been working but is not.

It's also easier and less disruptive to rename (or add in!) a named transition for tasks to follow if a given team's workflow is missing one of the opening transitions (eg "To Do" or "Backlog"). Changing a state's name is wildly disruptive, and requires also editing multiple board's column views, which is just crazy.

PR #54, opened 2015-11-02
Add a processor for Sentry to report failing URLs

singingwolfboy:
Sentry is reporting some exceptions from requests getting a status code 404, but isn't reporting on exactly *which* URLs are being requested. This processor should add that information. There's even a test!

@nedbat or @sarina, can one of you review this? [The `raven` documentation is here.](http://raven.readthedocs.org/) (`raven` is the name of the client software that reports clients to Sentry.)


sarina:
@smarnach FYI - i think this is related to what you were seeing

sarina:
in that, yes you were seeing some errors, but we don't even know what they were. now we'll have better reporting.

sarina:
I think this is good :+1: 

We'll look into adding validation to the process_pr function as well.

smarnach:
@sarina Good to see this change!  (And good to know where the code for the webhoook is – I'll read it on the next train journey, so I know what this thing is doing.)

sarina:
Yeah Sven we've made a few PRs this morning to explicitly restrict the
process_pr task to repos the bot has permission on.

smarnach:
@sarina That's even better than improving the error messages as I suggested.  If it isn't doing anything in the first place, there is little room for things going wrong. :)

PR #55, opened 2015-11-04
github_pr_repo and github_pr_num didn't realize they were getting res…

singingwolfboy:
…ponse objects


singingwolfboy:
@sarina, can you review this? It should resolve https://app.getsentry.com/edx-1/openedx-webhooks/group/88968953/

sarina:
OK, makes sense. :+1:

PR #58, opened 2015-11-04
Don't need to parse the JSON twice

singingwolfboy:
@sarina


sarina:
God I was so close to getthing this right. But so freaking far away. :+1:

PR #60, opened 2015-12-01
Clarify the AUTHORS file process.

agaylard:
Hi @sarina, I suggest the following text for edx-webhook.

* Have I understood the process correctly?
* Is the text suitable?

### DIFF
--- a/openedx_webhooks/templates/github_community_pr_comment.md.j2
+++ b/openedx_webhooks/templates/github_community_pr_comment.md.j2
@@ -19,20 +19,24 @@ GitHub pull request interface. As a reminder,
 {% endfilter %}
 
 
-{%- if not has_signed_agreement or not in_authors_file %}
+
 {% filter replace("\n", " ")|trim %}
-We can't start reviewing your pull request until you've
   {%- if not has_signed_agreement %}
-  submitted a [signed contributor agreement](http://open.edx.org/sites/default/files/wysiwyg/individual-contributor-agreement.pdf)
-  or indicated your institutional affiliation
-  {%- endif %}
-  {%- if not has_signed_agreement and not in_authors_file %}
-  and
+  We can't start reviewing your pull request until you've submitted a
+  [signed contributor agreement](http://open.edx.org/sites/default/files/wysiwyg/individual-contributor-agreement.pdf)
+  or indicated your institutional affiliation.
   {%- endif %}
   {%- if not in_authors_file %}
-  added yourself to the [AUTHORS](https://github.com/{{ repo }}/blob/master/AUTHORS) file
+  This PR can only be merged when you have added yourself to the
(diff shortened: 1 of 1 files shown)


edx-webhook:
Thanks for the pull request, @agaylard! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fopenedx-webhooks&number=60

openedx_webhooks/templates/github_community_pr_comment.md.j2
```diff
@@ -19,20 +19,24 @@ GitHub pull request interface. As a reminder,
...
   {%- endif %}
   {%- if not in_authors_file %}
-  added yourself to the [AUTHORS](https://github.com/{{ repo }}/blob/master/AUTHORS) file
+  This PR can only be merged when you have added yourself to the
+  [AUTHORS](https://github.com/{{ repo }}/blob/master/AUTHORS) file.
+  However, please wait for the results of the review process before
+  doing so.  When this PR is approved, then please add yourself to the
+  [AUTHORS](https://github.com/{{ repo }}/blob/master/AUTHORS) file in
+  a SEPARATE PR, and then reference it in this PR, after which BOTH
+  PRs can be merged.
```
sarina: I think this is overly complicated, and we can simplify it a bit. How about:

"You haven't added yourself to the [AUTHORS](https://github.com/{{ repo }}/blob/master/AUTHORS) file for this repo. Please create a new pull request to do so!"

I don't think the timing is really that terribly important, and I don't want this message to seem too overwhelming. What do you think?

sarina:
Thanks for doing this @agaylard

openedx_webhooks/templates/github_community_pr_comment.md.j2
agaylard: Yes, perfect.

sarina:
LGTM. @singingwolfboy would you check as well?

agaylard:
Please also check my logic -- those nested `if`- `endif` blocks are tricky...

singingwolfboy:
This generally looks good to me. However, can you fix the indentation in the section you've changed? As you said, lots of nested `if`s are tricky, and having appropriate indentation can make a big difference. (Let me know if you need more specific guidance in how to do that.)

agaylard:
OK, I've indented the lines I changed.  I've left the rest of the lines as they were (unindented).

singingwolfboy:
:+1:

PR #72, opened 2016-03-18
Remove the default internal coverletter for PRs (CS-25)

cpennington:
This removes the default internal PR cover-letter in favor of the Github PR template (baseline here: https://github.com/edx/cookiecutter-django-ida/blob/master/%7B%7Bcookiecutter.repo_name%7D%7D/.github/PULL_REQUEST_TEMPLATE.md), but allows repos to still add internal-only specific comments by adding a template inside the repo.

@nedbat @jbarciauskas @nasthagiri @ormsbee @andy-armstrong @pdesjardins: Thoughts/questions/objections?


edx-webhook:
## PR Cover Letter
- [ ] A detailed description of the changes in the body of the PR
- [ ] If available, a link to the relevant JIRA ticket in the body of the PR
  - `[PROJ-XXXX](https://openedx.atlassian.net/browse/PROJ-XXXX)`

## Sandbox
- [ ] Build a sandbox for your branch and add a link here

## Testing
- [ ] i18n
- [ ] RTL
- [ ] Unit, integration, acceptance tests as appropriate
- [ ] Analytics
- [ ] Performance
- [ ] Database migrations are backwards-compatible

## Reviewers
If you've been tagged for review, please check your corresponding box once you've given the :+1:.
- [ ] Code review #1
- [ ] Code review #2
- [ ] Docs review (required for UI strings/error messages)
- [ ] UX review
- [ ] Accessibility review
- [ ] Product review

## Documentation considerations
- [ ] Are there user visible changes? If so, list who is affected, and include a pointer to the relevant code.
- [ ] Should this be mentioned in the release notes? If so, make sure that you included a description.

## DevOps assistance
If it's needed, or even if you suspect it might be needed, do any/all these things as early as possible. Don't surprise devops right before merging! For example, touching envs/aws.py or or envs/common.py are dead giveaways that you'll need assistance when pushing your changes out. Things that require devops attention include, but are not limited to:
- configuration changes
- feature flags
- migrations
Ways to get devops attention, pick-and-choose whichever is most appropriate for your specific PR:
- [ ] Tag devops on your PR (`@edx/devops`), either for review (more specific to devops = better) or just a general FYI
- [ ] File a support ticket

FYI: Tag anyone who might be interested in this PR here.

## Post-review
- [ ] Squash commits

andy-armstrong:
:+1: Makes sense to me. It looks like you need to update the tests btw.

andy-armstrong:
@cpennington I'm adding a GitHub PR template to some of the FedX repos, and one thing that became apparent is that we no longer mention two things that internal teams should (almost) always do:
 - provide a link to the JIRA ticket
 - provide a link to a sandbox

I'm wondering if we should leave a default cover letter for repos that just requests these two things. If not, then I think we're taking a slight step backward from where we are today with the TNL pull request template.

What do you think?

cpennington:
@andy-armstrong good points. I'll add those.

andy-armstrong:
@cpennington On further reflection, it doesn't make sense to request sandboxes and JIRA tickets from all repos. I guess it will be up to each repo to add such a cover letter if they want it. In fact, now what this repo provides isn't a cover letter template, since we already have a cover letter. It is now additional actions required for either edX or OSPR contributors.

cpennington:
Right, but I think asking edX employees to tag w/ a JIRA ticket as standard practice is probably a good thing.

cpennington:
Actually, now that I think about it, I think the JIRA ticket part of the problem might be better solved elsewhere (because it'll be annoying having the bot ask you for a ticket number every time, even if you've already added one).

cpennington:
@andy-armstrong ok, tests should work now (I also added a string that identifies an internal coverletter as having been added by a bot. We may want it for the other things that the webhook bot adds).

openedx_webhooks/tasks/github.py
```diff
@@ -55,15 +58,18 @@ def pull_request_opened(pull_request, ignore_internal=True, check_contractor=Tru
                 repo=repo, num=num,
             ),
         )
-        comment = {
-            "body": github_internal_cover_letter(pr),
-        }
-        url = "/repos/{repo}/issues/{num}/comments".format(
-            repo=repo, num=num,
-        )
+        coverletter = github_internal_cover_letter(pr),
```
andy-armstrong: Can you update the docstring to match the new behavior (GitHub won't let me tag it directly). It looks like it has been wrong for a while as it says it ignores internal pull requests.

andy-armstrong:
:+1: once the docstring reflects the current functionality.

cpennington:
@nedbat: Can I get a quick review on this?

openedx_webhooks/tasks/github.py
```diff
@@ -20,14 +20,17 @@
 from openedx_webhooks.jira_views import get_jira_custom_fields
 
 
+COVERLETTER_MARKER = ":grey_exclamation: This message was added automatically by the [OpenEdX Bot](http://github.com/edx/openedx-webhooks)"
```
nedbat: We shouldn't use this for anything but cover letters, or the `has_internal_cover_letter` function could get confused.

BTW: Markdown supports HTML, so we can use HTML comment syntax to include text in comments that won't distract human readers.

There's a comment here:

nedbat:
:+1:

PR #81, opened 2016-09-23
Add EditorConfig configuration

gsong:
Adding cross-editor basic local configuration to override global settings

/cc @edx/ospr

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -11,9 +11,6 @@ insert_final_newline = true
 max_line_length = 80
 trim_trailing_whitespace = true
 
-[*.md]
-trim_trailing_whitespace = false
-
 [*.py]
 indent_size = 4
 max_line_length = 120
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,19 @@
...
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 2
+indent_style = space
+insert_final_newline = true
+max_line_length = 80
+trim_trailing_whitespace = true
+
+[*.md]
```
nedbat: There are no .md files in this repo.

.editorconfig
```diff
@@ -0,0 +1,19 @@
+# This file is for unifying the coding style for different editors and IDEs.
+# More information at http://EditorConfig.org
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 2
```
nedbat: What files do we indent with two spaces? Odd to have it as the default.

nedbat:
:+1:

PR #82, opened 2016-09-23
Add Heroku instructions

gsong:
Add instructions we have *so far* for development using the Heroku infrastructure.

/cc @edx/ospr

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -20,6 +20,9 @@ Make sure you've installed:
    *Hint*: See `runtime.txt`_ for the exact version
 -  `Heroku Command Line`_
 
+   All ``heroku`` commands can be performed through the Heroku web-based
+   dashboard as well, if you don't want to use the CLI.
+
 Set up
 ~~~~~~
 
@@ -70,7 +73,7 @@ difficult to ensure consistent experience for each developer. Instead,
 we utilize the `pipeline`_ facility offered by Heroku to handle our
 development needs.
 
-The general development cycle will be:
+The general development cycle is:
 
 Code → Deploy branch to staging → Test → Iterate
 
@@ -96,6 +99,17 @@ Deploy
 
 In most cases, you'll want to deploy by promoting from staging to
 production.
+
+**Prior to the promotion**, make sure all the changes have been merged
+to ``master``, and you've deployed the ``master`` branch successfully to
+staging.
+
+The general workflow is:
+
+Merge to ``master`` → Deploy ``master`` to staging → Test → Promote to
+production
+
+When you're ready to promote from staging to production:
 
 .. code:: sh
 
--- a/README.rst
+++ b/README.rst
@@ -17,7 +17,7 @@ Make sure you've installed:
 -  Python 2.7.x development environment (virtualenv strongly
    recommended)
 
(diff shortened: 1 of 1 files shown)


jbarciauskas:
@nedbat

README.rst
```diff
@@ -1,23 +1,178 @@
...
+
+The general development cycle is:
+
+Code → Deploy branch to staging → Test → Iterate
+
+To deploy a local branch to staging:
+
+.. code:: sh
+
+    git push heroku [branch_or_tag_or_hash:]master
```
nedbat: A handy command would be `git push heroku @:master` to push the current head to heroku's master.

README.rst
```diff
@@ -1,23 +1,178 @@
...
+
+Ensure the same versions are deployed
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Make sure the same git commit is deployed to both environments. First
+see what's deployed on staging:
+
+.. code:: sh
+
+    heroku releases -n 1
```
nedbat: This seems to need an app argument?
```
[:/src/edx/repo-tools/openedx-webhooks] [webhook] master ± heroku releases -n 1
 ▸    Error: No app specified
 ▸    Usage: heroku releases --app APP
 ▸    We don't know which app to run this on.
 ▸    Run this command from inside an app folder or specify which app to use with --app APP
 ▸
 ▸    https://devcenter.heroku.com/articles/using-the-cli#app-commands
```
Did you have the app specified implicitly somehow?

README.rst
```diff
@@ -1,23 +1,178 @@
...
+       heroku login
+
+   Authenticating is required to allow both the ``heroku`` and ``git``
+   commands to operate.
+
+2. Add the Heroku app repo as a git remote:
+
+   .. code:: sh
+
+       heroku git:remote -a openedx-webhooks-staging
```
gsong: I've added you and @shaunagm as members of this specific Heroku app.

README.rst
gsong: Yes. I believe most or all `heroku` commands look for the app referenced via `heroku` git remote as default. The reason it didn't work for you is probably because you didn't have view permissions to the `openedx-webhooks-staging` app.

README.rst
nedbat: Thanks, it works now.

README.rst
nedbat: My point is that there's a literal command that will always be what you want, rather than having a placeholder here that needs to be filled in with the name of the current branch.

README.rst
nedbat: Yup, now it works.

README.rst
gsong: I guess I see it differently. I'm not sure deploying from your current working branch is *always* the workflow.

If you feel strongly about it, I can hard code the instructions to `@` and rely on developers to know that they can deploy any revision reference they want.

nedbat: A good compromise would be to keep the line you have, and then say, "if you want to push your current code, use: `git push heroku @:master`"  Then they have both the general form, and the 95% convenience example.

nedbat:
:+1:

PR #83, opened 2016-09-23
Additional logging for GitHub PR web hook handler

gsong:
This represents the first part of the logging of our GitHub PR web hook handler. There will be a separate PR that handles the logging for the Celery tasks that perform the work.

In addition to the logging:

* Add a script to parse out results from Celery Redis backend
* Add sample web hook requests from GitHub for future reference (and possibly used for testing)

/cc @edx/ospr


openedx_webhooks/github_views.py
```diff
@@ -46,29 +49,34 @@ def pull_request():
...
     repo = pr["base"]["repo"]["full_name"].decode('utf-8')
     action = event["action"]
+    # `synchronize` action is when a new commit is made for the PR
     ignored_actions = set(("labeled", "synchronize"))
     if action in ignored_actions:
         msg = "Ignoring {action} events from github".format(action=action)
+        print(msg, file=sys.stderr)
         return msg, 200
 
+    _msg = "{}/pull/{} {}".format(repo, PR_NUMBER, action)
```
nedbat: An underscored local seems odd. And this isn't a message. Seems like "activity" to me.

openedx_webhooks/github_views.py
```diff
@@ -46,29 +49,34 @@ def pull_request():
...
     if action == "opened":
+        msg = "{}, processing…".format(_msg)
```
nedbat: I hope we won't regret non-ASCII in log messages later :)

openedx_webhooks/tasks/github.py
```diff
@@ -9,6 +9,7 @@
 from flask import render_template, render_template_string
 from urlobject import URLObject
 
+# TODO: Why aren't these relative imports?
```
nedbat: Should they be?

tests/payloads/github.headers.txt
```diff
@@ -0,0 +1,7 @@
+Request URL: https://openedx-webhooks.herokuapp.com/github/pr
```
nedbat: I don't see this file (or github.json) referenced anywhere, and there are no new tests added.  Are these for a different pull request?

openedx_webhooks/tasks/github.py
gsong: I would think so, since we're operating within the `openedx_webhooks` package.

tests/payloads/github.headers.txt
gsong: These are place holders for now. It was needed for me to reference as I unpack the file. We may use these as fixtures later. Or we may not, I'm not sure yet. But this is the raw payload that GitHub sends us.

openedx_webhooks/github_views.py
gsong: What do you think would be a better variable name? `activity`?

openedx_webhooks/github_views.py
```diff
@@ -46,29 +49,34 @@ def pull_request():
...
+        print(msg, file=sys.stderr)
         return msg, 200
 
+    pr_activity = "{}/pull/{} {}".format(repo, PR_NUMBER, action)
     if action == "opened":
+        msg = "{}, processing...".format(pr_activity)
+        print(msg, file=sys.stderr)
         result = pull_request_opened.delay(pr, wsgi_environ=minimal_wsgi_environ())
     elif action == "closed":
+        msg = "{}, processing…".format(pr_activity)
```
nedbat: We should ascii-fy this one also.

openedx_webhooks/github_views.py
```diff
@@ -33,10 +33,13 @@ def pull_request():
...
     """
+    # TODO: We need to untangle this, there are **four** `return`s in this function!
     try:
         event = request.get_json()
     except ValueError:
-        raise ValueError("Invalid JSON from Github: {data}".format(data=request.data))
+        msg = "Invalid JSON from Github: {data}".format(data=request.data)
+        print(msg, file=sys.stderr)
+        raise ValueError(msg)
     sentry.client.extra_context({"event": event})
```
nedbat: I think it would be useful right here to add a log line that an event was received.  It will help us diagnose problems if something goes wrong before one of the exiting log lines is reached.

openedx_webhooks/github_views.py
```diff
@@ -46,29 +49,34 @@ def pull_request():
         return "PONG"
 
     pr = event["pull_request"]
+    PR_NUMBER = pr['number']
```
nedbat: Shouldn't this be `pr_number`? It's not a constant.

gsong: It's a dynamically valued constant within the scope of the function, since it's *not* to be changed for the duration of the function execution. Again, if you feel strongly about it, I can change it.

nedbat: I've never seen something like this before, and I don't understand: pr_activity is value that doesn't change during the function, but you gave it a lowercase name.  This isn't a definition of constant that other programmers are going to get.

gsong: I changed it. But FYI: <https://en.wikipedia.org/wiki/Constant_(computer_programming)#Dynamically-valued_constants>.

Also, since Python seems to be taking a lot of cues from ES6: <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const>

nedbat: Pointing to other languages' features won't change my mind. :) This doesn't make any sense in Python. You're the first person I've seen do it, and it will confuse everyone.

gsong: I am curious to find out why you don't think it makes any sense in Python. Always open to learning.

nedbat: I thought about this while swimming :)

# Why C has const

C provides for pointers to variables (and C++ has references). This means that variables can be affected by code far from their declaration.  I can pass a reference to a variable into a function, and the function can operate on the variable. The const declaration provides a way to prevent modification at a distance.

# Why JavaScript has const

I don't know why JavaScript has const. A five-minute search didn't turn up a good answer, just a lot of people pointing out that const is different than immutability. I suspect const is useful to the compiler or JIT engines, since the declaration provides information about the future that can guide their work.

# Why Python doesn't have const

A few reasons:

1. Unlike C, Python has no way to make a reference to a name, so there's no risk that a function will suprisingly change a variable passed into it.

2. Unlike JavaScript, Python has no declarations that could take a const modifier, so there's no way to provide guidance to the compiler or JIT.

3. Because Python only has naming conventions, and no declarations enforced by the language, we are relying on the name to guide the developer in their work.  This is very different than languages with a true const declaration that can be enforced in whatever way the language does.


# Semantic problems with "dynamically valued constants"

It seems like `PR_NUMBER = pr['number']` is stating that we are never going to change PR_NUMBER after this point.  Suppose I realize later that the value I got is a string, and I'd really find an integer more convenient.  Am I allowed to write this?

```python 
PR_NUMBER = pr['number']
PR_NUMBER = int(PR_NUMBER)
```

Seems like that would be wrong, since I am reassigning that name.  So can I write it like this?

```python
PR_NUMBER = int(pr['number'])
```

Now I've only assigned it once, so it's a constant. But that's a very subtle distinction between those two snippets, some might even say a trivial distinction.  Do we really want to be switching between lower-case and upper-case based on the number of assignment statements? It's an uninteresting consideration.

A constant is a thing whose value never changes.  "Dynamically-valued constants" is literally an oxymoron.

In the code in question, why was PR_NUMBER a "constant", but not other names?  By my reading of the code, we have eight other names with only a single assignment statement (event, pr, repo, action, ignored_actions, pr_activity, status_url, resp).   Why aren't they constants?

And then we get to the thornier issue of `result`, which is assigned in two different places, but only one of them will be executed in any given run of the function. Is that a constant?

Lastly, suppose we never reassign one of the parameters to a function? Should the parameter be named with all upper-case?  That would leak internal details to the callers of the function, and that detail could change as the function is further developed.

Overall, it seems to me that the number of times you assign to a name is an uninteresting detail, unless it can help the compiler or JIT.  Naming it with all upper-case letters in Python won't help the compiler or JIT.


# Why does Python have constants at all?

OK, so after all that, why use upper-case constants in Python at all? They aren't enforced, and you can still get into pedantic debates about multiple assignments as a way to construct the value.

First, true constants, names which refer to values which are always the same, every time you refer to them, are very useful in understanding a program. That a value never changes is an interesting detail to understand about a value.

Second, module-level or class-level constants are names that will be used across large bodies of code, far from where they were created.  Communicating the const-ness with a name is helpful to understanding those distant values.  Names within a single function don't have that same understanding-at-a-distance issue.

gsong: 👍 re: *understanding-at-a-distance*

Thank you for taking the time to think this through, @nedbat. And `const` in JS does not equate to immutability, but is rather about reassignment of the variable. There's a debate in the JS community about the need to uppercase those names as well.

nedbat: Go tell the JavaScript people that it would be silly to upper-case the names. Tell them I said so. :)

gsong: Agreed, especially since there's syntax for explicit declaration of `const` vs. `let`.

nedbat:
:+1: Merge it!

nedbat:
But first squash the commits?  Have we talked about this before?

nedbat:
Or don't squash the commits :)

PR #84, opened 2016-09-28
Update requirements

gsong:
@edx/ospr


nedbat:
I'm new to this pip-tools world.  How often do we think we'll have PRs like this? The requirements were last updated three weeks ago.

gsong:
It's unlikely we'll have just updates for updates sake. In my previous migration to pip-tools, I didn't touch the versions at all. So this represents an update for this chuck of "major" work.

I think in general we should update packages when we do feature updates.

nedbat:
OK, :+1: I guess. I don't know how I would decide to not approve this pull request.

PR #85, opened 2016-09-28
Additional logging for GitHub PR web hook Celery tasks

gsong:
Add logging for Celery tasks which are used to handle GitHub PR open and close events.

Also:
* Make sure Heroku Celery worker task actually outputs logging statements
* Created an example Celery task to test logging and establish testing pattern

/cc @edx/ospr

### DIFF
--- a/openedx_webhooks/tasks/example.py
+++ b/openedx_webhooks/tasks/example.py
@@ -1,0 +1,38 @@
+# -*- coding: utf-8 -*-
+"""
+Example Celery task, can be used for testing the Celery infrastructure
+at Heroku.
+
+Example usage::
+
+    from openedx_webhooks.tasks.example import add
+    add.delay(1, 2)
+"""
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+from openedx_webhooks import celery
+from openedx_webhooks.tasks.utils import log_info
+
+celery.conf.update({
+    'BROKER_URL': 'ironmq://',
+    'CELERY_TASK_SERIALIZER': 'json',
+    'CELERY_RESULT_SERIALIZER': 'json',
+})
+
+
+@celery.task(name='example.add', bind=True)
+def add(self, num1, num2):
+    """
+    Example Celery task.
+
+    Args:
+        num1 (int): First number
+        num2 (int): Second number
+
+    Returns:
+        int: Sum
+    """
+    result = num1 + num2
+    msg = "Results for `{}`: {}".format(self.name, result)
+    log_info(self.request, msg)
+    return result
--- a/openedx_webhooks/tasks/github.py
+++ b/openedx_webhooks/tasks/github.py
@@ -1,31 +1,35 @@
 # coding=utf-8
-from __future__ import unicode_literals, print_function
+from __future__ import absolute_import, division, print_function, unicode_literals
 
+from datetime import date
 import json
 import re
-from datetime import date
 
-from iso8601 import parse_date
 from flask import render_template, render_template_string
+from iso8601 import parse_date
 from urlobject import URLObject
+import requests
 
 # TODO: Why aren't these relative imports?
 from openedx_webhooks import sentry, celery
-from openedx_webhooks.tasks import logger
-from openedx_webhooks.oauth import github_bp, jira_bp
 from openedx_webhooks.info import (
-    get_people_file, is_internal_pull_request, is_contractor_pull_request,
+    get_people_file,
     is_beta_tester_pull_request,
+    is_contractor_pull_request,
+    is_internal_pull_request,
 )
-from openedx_webhooks.utils import memoize, paginated_get
 from openedx_webhooks.jira_views import get_jira_custom_fields
+from openedx_webhooks.oauth import github_bp, jira_bp
+from openedx_webhooks.tasks import logger
+from openedx_webhooks.tasks.utils import log_error, log_info, log_request_response
+from openedx_webhooks.utils import memoize, paginated_get
 
 
 COVERLETTER_MARKER = "<!-- open edx coverletter -->"
 
 
-@celery.task
-def pull_request_opened(pull_request, ignore_internal=True, check_contractor=True):
+@celery.task(bind=True)
+def pull_request_opened(self, pull_request, ignore_internal=True, check_contractor=True):
     """
     Process a pull request. This is called when a pull request is opened, or
     when the pull requests of a repo are re-scanned. By default, this function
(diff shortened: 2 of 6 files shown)


openedx_webhooks/tasks/example.py
```diff
@@ -0,0 +1,40 @@
...
+at Heroku.
+
+Example usage::
+
+    from openedx_webhooks.tasks.example import add
+    add.delay(1, 2)
+"""
+from __future__ import (
+    absolute_import, division, print_function, unicode_literals
+)
```
nedbat: If we have to put boilerplate `__future__` imports that we don't need at the top of the file, can we at least put it on one line instead of three?

openedx_webhooks/tasks/github.py
```diff
@@ -1,30 +1,34 @@
...
+from datetime import date
 import json
 import re
-from datetime import date
 
-from iso8601 import parse_date
 from flask import render_template, render_template_string
+from iso8601 import parse_date
 from urlobject import URLObject
+import requests
```
nedbat: Shouldn't this be before urlobject?

openedx_webhooks/tasks/utils.py
```diff
@@ -0,0 +1,118 @@
...
+    log_info(task_request, msg)
+
+
+def log_request_response(task_request, response):
+    """
+    Logs two messages using the Celery's app logger at the INFO level for
+    request and response.
+
+    Args:
+        task_request (Task.request): Celery task instance `request`_ object.
```
nedbat: These lines duplicate information. Why give the type twice? And as dutiful as it is, we don't need the links in every docstring.  If you want to provide the link, put it in the module docstring.  As it is, there's so much repetitive docstring boilerplate, it's hard to see the code.

tests/tasks/test_example.py
```diff
@@ -0,0 +1,11 @@
+import pytest
+
+from openedx_webhooks.tasks.example import add
+
+
+def test_add():
+    assert add(5, 6) == 11
+
+
+if __name__ == '__main__':
```
nedbat: We don't need this in every test file, do we? I'm not a heavy pytest user, but I would have thought it could take care of this for us.

openedx_webhooks/tasks/github.py
```diff
@@ -1,30 +1,34 @@
...
+    get_people_file,
     is_beta_tester_pull_request,
+    is_contractor_pull_request,
+    is_internal_pull_request,
 )
-from openedx_webhooks.utils import memoize, paginated_get
 from openedx_webhooks.jira_views import get_jira_custom_fields
+from openedx_webhooks.oauth import github_bp, jira_bp
+from openedx_webhooks.tasks import logger
+from openedx_webhooks.tasks.utils import log_error, log_info, log_request_response
```
nedbat: I'd have though the tasks package was just for tasks, and that utils for tasks could go into the existing utils module, but I guess people differ on that.

openedx_webhooks/tasks/github.py
```diff
@@ -242,23 +252,25 @@ def pull_request_closed(pull_request):
                 valid=", ".join(t["to"]["name"].decode('utf-8') for t in transitions),
             )
         )
+        log_error(self.request, fail_msg)
         raise Exception(fail_msg)
 
+    log_info(self.request, 'Chaing JIRA issue status...')
```
nedbat: Typo

openedx_webhooks/tasks/github.py
```diff
@@ -43,6 +47,7 @@ def pull_request_opened(pull_request, ignore_internal=True, check_contractor=Tru
     element in the tuple is a boolean indicating if this function did any
     work, such as making a JIRA issue or commenting on the pull request.
     """
```
nedbat: A "starting task" log message would be helpful here.

openedx_webhooks/tasks/github.py
```diff
@@ -53,46 +58,43 @@ def pull_request_opened(pull_request, ignore_internal=True, check_contractor=Tru
...
 
     if is_internal_pr and not has_cl and is_beta:
-        logger.info(
-            "Adding cover letter template to PR #{num} against {repo}".format(
-                repo=repo, num=num,
-            ),
-        )
+        msg = "Adding cover letter template to PR #{num} against {repo}".format(repo=repo, num=num)
+        log_info(self.request, msg)
         coverletter = github_internal_cover_letter(pr),
```
nedbat: That trailing comma looks suspicious.....

openedx_webhooks/tasks/github.py
```diff
@@ -53,46 +58,43 @@ def pull_request_opened(pull_request, ignore_internal=True, check_contractor=Tru
...
             comment = {
                 "body": coverletter
             }
-            url = "/repos/{repo}/issues/{num}/comments".format(
-                repo=repo, num=num,
-            )
+            url = "/repos/{repo}/issues/{num}/comments".format(repo=repo, num=num)
 
             comment_resp = github.post(url, json=comment)
+            log_request_response(self.request, comment_resp)
```
nedbat: Is this log message enough to know what happened? I would have expected a "Added cover letter" message.

openedx_webhooks/tasks/github.py
gsong: I usually sort `from` imports before `import`. Is it convention here to just sort by package name? Actually most code I've seen doesn't sort at all.

openedx_webhooks/tasks/github.py
gsong: I personally prefer to keep the code as close to modules that use it, and only refactor up the tree if other modules up the tree start using them. Makes for more modules, but I think it's more clear from an organizational standpoint.

openedx_webhooks/tasks/github.py
nedbat: OK, I understand your convention.

openedx_webhooks/tasks/github.py
gsong: I think so. We would have two log statements, one request to GitHub and one response from GitHub. In the response, we're logging status code and reason, so we should see `201 Created` if all is well, or an error code if something went wrong.

openedx_webhooks/tasks/github.py
gsong: What was I typing!?

tests/tasks/test_example.py
gsong: My instinct was the same, but again, it wasn't clear to me what the edX standards are. I thought I would err on the side of more formal documentation.

openedx_webhooks/tasks/utils.py
gsong: My instinct was the same, but again, it wasn't clear to me what the edX standards are. I thought I would err on the side of more formal documentation.

tests/tasks/test_example.py
gsong: No, I can take it out. The idea is that you can just execute the file itself and the test will run.

openedx_webhooks/tasks/github.py
gsong: Mucho.

nedbat:
:+1: squash and merge :)

PR #89, opened 2016-10-26
Add tests for new code

gsong:
Fairly complete unit test coverage of the new code.

`make test` to see all the test pass! 🎉

/cc @edx/ospr

### DIFF
--- a/openedx_webhooks/github/dispatcher/rules/github_activity.py
+++ b/openedx_webhooks/github/dispatcher/rules/github_activity.py
@@ -6,17 +6,17 @@ Update JIRA issue with latest GitHub activity.
 from __future__ import absolute_import, division, print_function, unicode_literals
 
 from ....jira.tasks import update_latest_github_activity
-from ....lib.jira import jira
+from ....lib.jira.decorators import inject_jira
 from ...models import GithubEvent
 
 
-def _find_issues_for_pull_request(pull_request_url, jira=jira):
+def _find_issues_for_pull_request(jira, pull_request_url):
     """
     Find corresponding JIRA issues for a given GitHub pull request.
 
     Arguments:
+        jira (jira.JIRA): An authenticated JIRA API client session
         pull_request_url (str)
-        jira (jira.JIRA)
 
     Returns:
         jira.client.ResultList[jira.Issue]
@@ -25,18 +25,20 @@ def _find_issues_for_pull_request(pull_request_url, jira=jira):
     return jira.search_issues(jql)
 
 
-def _process(event_type, raw_event):
+def _process(jira, event_type, raw_event):
     """
     Update each corresponding JIRA issue with GitHub activity.
 
     Arguments:
+        jira (jira.JIRA): An authenticated JIRA API client session
         event_type (str): GitHub event type
         raw_event (Dict[str, Any]): The parsed event payload
     """
     event = GithubEvent(event_type, raw_event)
-    issues = _find_issues_for_pull_request(event.html_url)
+    issues = _find_issues_for_pull_request(jira, event.html_url)
     for issue in issues:
         update_latest_github_activity(
+            jira,
             issue.id,
             event.description,
             event.user.login,
@@ -45,11 +47,13 @@ def _process(event_type, raw_event):
         )
 
 
-def run(event_type, raw_event):
+@inject_jira
+def run(jira, event_type, raw_event):
     """
     Process the incoming event.
 
     Arguments:
+        jira (jira.JIRA): An authenticated JIRA API client session
         event_type (str): `GitHub event type`_
         raw_event (Dict[str, Any]): The parsed event payload
 
@@ -63,4 +67,4 @@ def run(event_type, raw_event):
         'pull_request_review_comment',
     )
     if event_type in event_types:
-        _process(event_type, raw_event)
+        _process(jira, event_type, raw_event)
--- a/openedx_webhooks/lib/edx_repo_tools_data/models.py
+++ b/openedx_webhooks/lib/edx_repo_tools_data/models.py
@@ -29,7 +29,7 @@ class People(object):
             data (Dict[str, Any]): The raw dictionary as parsed from
                 the yaml file
         """
-        self.data = data
+        self._data = data
 
     def get(self, key):
         """
@@ -45,7 +45,7 @@ class People(object):
             openedx_webhooks.lib.exceptions.NotFoundError: If key cannot
                 be found
         """
-        person = self.data.get(key)
+        person = self._data.get(key)
 
         if not person:
             raise NotFoundError("{} could not be found".format(key))
@@ -63,7 +63,6 @@ class Person(object):
             yaml file
     """
 
-    # TODO: Test
     def __init__(self, login, data):
         """
         Init.
@@ -74,14 +73,21 @@ class Person(object):
                 the yaml file
         """
         self.login = login
-        self.data = data
+        self._data = data
+
+    @property
+    def _before(self):
+        """
+        Optional[Dict[str, Dict]]: User's `before` data.
+        """
+        return self._data.get('before')
 
     @property
     def agreement(self):
         """
         Optional[str]: User's agreement.
         """
-        data = self.data.get('agreement')
+        data = self._data.get('agreement')
         if data == 'none':
             data = None
         return data
@@ -97,18 +103,11 @@ class Person(object):
         return expired
 
     @property
-    def before(self):
-        """
-        Optional[Dict[str, Dict]]: User's `before` data.
-        """
-        return self.data.get('before')
-
-    @property
     def agreement_expires_on(self):
         """
-        Optional[datetime.date]: User's `before` data.
+        Optional[datetime.date]: When did the user's agreement expire?
         """
-        expires_on = self.data.get('expires_on')
+        expires_on = self._data.get('expires_on')
         if expires_on:
             return expires_on
 
@@ -119,8 +118,8 @@ class Person(object):
             yesterday = arrow.now().replace(days=-1).date()
             expires_on = yesterday
 
-        if not self.agreement and self.before:
-            expires_on = max(self.before.keys())
+        if not self.agreement and self._before:
+            expires_on = max(self._before.keys())
 
         return expires_on
 
@@ -129,7 +128,7 @@ class Person(object):
         """
         Optional[str]: User's institution.
         """
-        data = self.data.get('institution')
+        data = self._data.get('institution')
         return self.agreement and data
 
     @property
--- a/openedx_webhooks/lib/edx_repo_tools_data/test/models/conftest.py
+++ b/openedx_webhooks/lib/edx_repo_tools_data/test/models/conftest.py
@@ -1,0 +1,82 @@
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+import datetime
+
+import pytest
+
+from openedx_webhooks.lib.edx_repo_tools_data.models import People, Person
+
+
+@pytest.fixture
+def active_data():
+    data = {'active-person': {
+        'agreement': 'individual',
+    }}
+    return data
+
+
+@pytest.fixture
+def expired_data():
+    data = {'expired-person': {
+        'agreement': 'institution',
+        'expires_on': datetime.date(2012, 10, 1),
+        'institution': 'edX',
+    }}
+    return data
+
+
+@pytest.fixture
+def active_person(active_data):
+    k, v = list(active_data.items())[0]
+    return Person(k, v)
+
+
+@pytest.fixture
+def active_edx_person():
+    person = Person('active-person', {
+        'agreement': 'institution',
+        'institution': 'edX',
+    })
+    return person
+
+
+@pytest.fixture
+def active_non_edx_person():
+    person = Person('active-person', {
+        'agreement': 'institution',
+        'institution': 'Shield',
+    })
+    return person
+
+
+@pytest.fixture
+def expired_person(expired_data):
+    k, v = list(expired_data.items())[0]
+    return Person(k, v)
+
+
+@pytest.fixture
+def before():
+    data = {
+        datetime.date(2016, 1, 9): {},
+        datetime.date(2016, 7, 29): {},
+        datetime.date(2016, 8, 8): {},
+    }
+    return data
+
+
+@pytest.fixture
+def before_expired_person(before):
+    person = Person('expired-before-person', {
+        'agreement': 'none',
+        'before': before,
+    })
+    return person
+
+
+@pytest.fixture
+def people(active_data, expired_data):
+    data = active_data.copy()
+    data.update(expired_data)
+    return People(data)
--- a/openedx_webhooks/lib/edx_repo_tools_data/test/models/test_people.py
+++ b/openedx_webhooks/lib/edx_repo_tools_data/test/models/test_people.py
@@ -1,0 +1,16 @@
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+import pytest
+
+from openedx_webhooks.lib.exceptions import NotFoundError
+
+
+class TestGet:
+    def test_get(self, people):
+        expected = 'active-person'
+        assert people.get(expected).login == expected
+
+    def test_miss(self, people):
+        with pytest.raises(NotFoundError):
+            people.get('foo')
--- a/openedx_webhooks/lib/github/test/models/test_github_webhook_event.py
+++ b/openedx_webhooks/lib/github/test/models/test_github_webhook_event.py
@@ -1,0 +1,81 @@
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import, division, print_function, unicode_literals
+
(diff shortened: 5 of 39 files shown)


openedx_webhooks/lib/edx_repo_tools_data/test/models/conftest.py
```diff
@@ -0,0 +1,84 @@
...
+        'expires_on': datetime.date(2012, 10, 1),
+        'institution': 'edX',
+    }}
+    return data
+
+
+@pytest.fixture
+def active_person(active_data):
+    for k, v in active_data.items():
+        person = Person(k, v)
```
nedbat: This looks like a mistake: it will loop over "all" of active_data, but only make one Person.  I understand that active_data only has one item, but looping is odd here. We need some acknowledgement here of what is going on.  You don't want popitem because that will mutate the dict. Maybe `k, v = next(active_data.items())`, with a comment about there being only one item in active_data? Or maybe an assert on the length of active_data?

openedx_webhooks/lib/edx_repo_tools_data/models.py
```diff
@@ -125,11 +117,18 @@ def agreement_expires_on(self):
         return expires_on
 
     @property
+    def before(self):
+        """
+        Optional[Dict[str, Dict]]: User's `before` data.
+        """
+        return self._data.get('before')
```
nedbat: I thought we would prefer to keep the "before" details internal, so as not to further expose the before/expired confusion.

openedx_webhooks/lib/edx_repo_tools_data/test/models/test_people.py
```diff
@@ -0,0 +1,16 @@
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+import pytest
+
+from openedx_webhooks.lib.exceptions import NotFoundError
+
+
+class TestGet:
```
nedbat: Usually once people have drunk the pytest kool-aid with fixtures, etc, they no longer write test classes.  Why do you prefer to?

openedx_webhooks/lib/github/test/models/test_github_webhook_event.py
```diff
@@ -0,0 +1,83 @@
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+import arrow
+import pytest
+
+from openedx_webhooks.lib.github.models import GithubWebHookEvent
```
nedbat: Why not a relative import here?

openedx_webhooks/lib/github/test/models/test_github_webhook_event.py
```diff
@@ -0,0 +1,83 @@
...
+
+def test_githubwebhookevent_sender_login(issue_comment):
+    assert issue_comment.sender_login == 'issue-sender'
+
+
+class TestGithubWebHookEventUpdatedAt:
+    def test_issue(self, issue_comment):
+        assert (
+            issue_comment.updated_at ==
+            arrow.get('2016-10-24T18:53:10Z').datetime
```
nedbat: I'm a little concerned about using arrow.get in the test.  Better would be to use the raw datetime constructor to be absolutely sure we are getting the datetime we need.  It's not happening here, but if we use "arrow.get" in both the product code and the test code, there's a risk that we use it wrong in both places, and the test confirms the product output, but they are both wrong.

openedx_webhooks/lib/github/test/utils/test_get_most_recent_hook.py
```diff
@@ -0,0 +1,74 @@
...
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+from collections import namedtuple
+
+import arrow
+import pytest
+
+from openedx_webhooks.lib.github.utils import _get_most_recent_hook
+
+Hook = namedtuple('Hook', ['active', 'updated_at'])
```
nedbat: It's confusing to have two different stub Hook namedtuples in two different files.  Could we have a single stub class to use everywhere?

openedx_webhooks/lib/github/test/utils/test_repo_contains_webhook.py
```diff
@@ -0,0 +1,26 @@
...
+from openedx_webhooks.lib.github.utils import repo_contains_webhook
+
+
+class TestRepoContainsWebhook:
+    def test_include_inactive(self, repo):
+        result = repo_contains_webhook(repo, 'http://test.inactive')
+        assert result is True
+
+    def test_no_inactive(self, repo):
+        result = repo_contains_webhook(repo, 'http://test.inactive', True)
```
nedbat: I'd use an explicit keyword with a boolean argument.

openedx_webhooks/lib/github/utils.py
```diff
@@ -42,7 +41,7 @@ def repo_name(repo):
     return "{0.owner.login}/{0.name}".format(repo)
 
 
-def create_hook(repo, config, events, active, dry_run=False):
+def create_hook(repo, config, events, active, dry_run=False):  # pragma: no cover
```
nedbat: We shouldn't add pragmas like this if we'd like to eventually test this code.

openedx_webhooks/lib/jira/client.py
```diff
@@ -0,0 +1,24 @@
...
+    access_token=os.environ.get('JIRA_ACCESS_TOKEN'),
+    access_token_secret=os.environ.get('JIRA_ACCESS_TOKEN_SECRET'),
+    consumer_key=os.environ.get('JIRA_OAUTH_CONSUMER_KEY'),
+    key_cert=b64decode(os.environ.get('JIRA_OAUTH_PRIVATE_KEY')),
+)
+
+jira = JIRA(_server, oauth=_oauth_info)
+"""
+(jira.JIRA): An authenticated JIRA API client session
+"""
```
nedbat: Did we discuss this already?  Why isn't this a comment?

openedx_webhooks/lib/jira/test/conftest.py
```diff
@@ -0,0 +1,31 @@
...
+from openedx_webhooks.lib.jira.models import JiraField, JiraFields
+
+
+@pytest.fixture
+def datum():
+    return {'name': 'test01', 'first': True, 'id': 'id_test01'}
+
+
+@pytest.fixture
+def data(datum):
```
nedbat: "data" seems like a really generic name for a fixture.

openedx_webhooks/lib/jira/test/test_lib_jira_models.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class TestJiraField:
+    def test_getattr(self, field):
+        assert field.name == 'test01'
+        assert field.first is True
+
+    def test_getattr_error(self, field):
+        with pytest.raises(AttributeError):
+            field.nope
```
nedbat: What is this testing?

.coveragerc
```diff
@@ -1,3 +1,13 @@
...
+omit =
+    */conftest.py
+    */test_*.py
+    openedx_webhooks/github/dispatcher/rules/github_activity.py
+    openedx_webhooks/github/models.py
+    openedx_webhooks/jira/tasks.py
+    openedx_webhooks/lib/edx_repo_tools_data/utils.py
+    openedx_webhooks/lib/jira/client.py
+    openedx_webhooks/webhook_confs.py
+    openedx_webhooks/worker.py
```
gsong: None of these modules are very interesting, they fall into one of two categories:

* They are façades to third party API clients
* They call on other tested modules

I think at the time that they do more heavy lifting, we can remove the exceptions so they don't report false positives in terms of needing test coverage.

openedx_webhooks/github/dispatcher/rules/github_activity.py
```diff
@@ -21,6 +19,8 @@ def _find_issues_for_pull_request(pull_request_url, jira=jira):
     Returns:
         jira.client.ResultList[jira.Issue]
     """
+    if jira is None:
+        from ....lib.jira.client import jira
```
gsong: The JIRA client relies on environment variables to be set, which may not be present during testing (and don't need to be, since we're not actually making calls to the JIRA server).

The problem is the JIRA client fails if it can't make a proper connection to the server.

openedx_webhooks/github/dispatcher/rules/github_activity.py
```diff
@@ -33,6 +33,7 @@ def _process(event_type, raw_event):
         event_type (str): GitHub event type
         raw_event (Dict[str, Any]): The parsed event payload
     """
+    from ....jira.tasks import update_latest_github_activity
```
gsong: See <https://github.com/edx/openedx-webhooks/pull/89#discussion_r85806545>.

openedx_webhooks/lib/edx_repo_tools_data/models.py
gsong: I'd rather defer this discussion until later, when we are designing the API to `repo-tools-data`. I'm not sure what the unifying interface is to façade both `before` and `expires_on`, since I'm not aware of all the use cases.

At least in this iteration, you only have to ask `Person.agreement_expired` and `Person.agreement_expires_on`.

We can always make `before` a private property.

openedx_webhooks/lib/edx_repo_tools_data/test/models/test_people.py
gsong: Test classes are just used for organization, right? In this case it makes sense to organize all the tests having to do with different scenarios of the same method. I'm anticipating that `People` will acquire more properties and methods.

openedx_webhooks/lib/github/test/models/test_github_webhook_event.py
gsong: Tests aren't organized in packages, so I don't think relative imports make sense.

.coveragerc
nedbat: I'm skeptical that we have code that is important to have, but uninteresting to test.  Excluding code from test coverage seems like a dangerous precedent.

openedx_webhooks/github/dispatcher/rules/github_activity.py
nedbat: I guess I would deal with that problem closer to where the environment variables are being read, rather than expecting the users of the code to remember that they need to make local imports.

openedx_webhooks/lib/edx_repo_tools_data/models.py
nedbat: In the past, the interface we needed to expose publically was, "what agreement was in place on a particular date".  "Agreement_expired" could be built on that, or alongside it.  I don't know of a time we needed "agreement_expires_on".

openedx_webhooks/lib/edx_repo_tools_data/test/models/test_people.py
nedbat: OK, it just seems odd to mix the cultures like this. There are places where you have test classes and test functions intermixed, where I can't see the logic.  With unittest, test classes are necessary, and are the basis for setUp methods, test case attributes, and so on.  These classes are giving us none of that.

openedx_webhooks/lib/github/test/models/test_github_webhook_event.py
nedbat: I would rather have a uniform way to import GithubWebHookEvent throughout the code base. As it is, we have absolute here, and relative in the product code.

openedx_webhooks/lib/github/test/utils/test_get_most_recent_hook.py
gsong: I think it's preferable to stub as little as possible. In this case, the two `Hook`s have different shapes, so I'm OK with them being only local to where they're used.

openedx_webhooks/lib/github/test/models/test_github_webhook_event.py
nedbat: This doesn't seem so bad to me: `datetime.datetime(2016, 10, 24, 18, 53, 10, tzinfo=pytz.UTC)`.

My concern is about duplicating a bit of the product code in the test code.  I guess it's not a big deal in this case, but you can see how that would be a risk.

nedbat: To stick with packages we already have, it would be `datetime.datetime(2016, 10, 24, 18, 53, 10, tzinfo=dateutil.tz.tzutc())`

openedx_webhooks/lib/github/utils.py
gsong: I'm not sure what we would be testing here.

openedx_webhooks/lib/jira/client.py
gsong: <http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html>, see `module_level_variable2`.

openedx_webhooks/lib/github/utils.py
nedbat: You've got 7 or 8 lines of code here, including two conditionals, string formatting, and exception raising. You would test that all that works. You have tests of GithubWebHookEvent, and all of its methods are one or two lines with no conditionals. I don't understand why that code is interesting to test, and this code is not?

openedx_webhooks/lib/jira/client.py
nedbat: It seems like they also allow a comment preceding the variable, which seems much more natural to me.

openedx_webhooks/lib/jira/test/test_lib_jira_models.py
gsong: That we fulfill the contract of the method and raise an appropriate error when the `JiraField` object doesn't contain an attribute.

openedx_webhooks/lib/edx_repo_tools_data/models.py
gsong: Can we agree to table the API design for another exercise?

openedx_webhooks/lib/github/test/models/test_github_webhook_event.py
gsong: I could put test modules alongside the modules they're testing, then we won't need the `test/` subdirectory (which isn't a package, so I literally *can't* do relative imports).

If we had these tests in a project `/tests` directory, we would have to do absolute imports as well, so I guess I don't see the problem.

openedx_webhooks/lib/edx_repo_tools_data/models.py
nedbat: I was answering your uncertainty about what use cases. We can defer the API design, though I would default to leaving the property private until we know we need it, rather than making it public and then later figuring out if we should take it out.

openedx_webhooks/lib/github/test/models/test_github_webhook_event.py
nedbat: Maybe it's not a big problem, but it seems unfortunate that in some places we have
```
from ..lib.github.models import GithubWebHookEvent
```
and in other places we have
```
from openedx_webhooks.lib.github.models import GithubWebHookEvent
```

nedbat:
Somewhere in here we were discussing relative vs absolute imports. I can't find any of it any more. What happened to it? :(

gsong:
@nedbat [Relative vs. absolute import discussion](https://github.com/edx/openedx-webhooks/pull/89#discussion_r85720860)

openedx_webhooks/lib/github/test/models/test_github_webhook_event.py
gsong: This is how I think of it: Use relative imports when files are intra-package, absolute imports otherwise. Tests are not part of the package, so should use absolute imports.

nedbat: ok

openedx_webhooks/lib/jira/decorators.py
```diff
@@ -0,0 +1,29 @@
...
+
+from functools import wraps
+
+from jira import JIRA
+
+
+def inject_jira(f):
+    """
+    Inject authenticated JIRA client (``jira``) if one is not supplied.
+    """
```
nedbat: This could use more explanation: the first argument to the function has to be the jira argument.

gsong:
@nedbat I believe all concerns have been addressed, let me know if this is a 👍.

openedx_webhooks/test/test_is_valid_payload.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+@pytest.fixture
+def good_signature(secret1, payload):
+    return _signature(secret1, payload)
+
+
+@pytest.fixture
+def bad_signature(secret2, payload):
+    return _signature(secret2, payload)
```
nedbat: This isn't a bad signature. It's just not the signature you expect in a test further down. This fixture is only used in that test. It'd be easier to understand the test if you move this into that test function, and call it wrong_signature instead.

gsong:
@nedbat Latest feedbacks have been addressed.

nedbat:
:+1:

PR #90, opened 2016-11-04
Update JIRA GitHub activity field names

gsong:
To be in alignment with the final implementation in production JIRA.

/cc @edx/ospr

### DIFF
--- a/openedx_webhooks/jira/tasks.py
+++ b/openedx_webhooks/jira/tasks.py
@@ -10,6 +10,11 @@ from ..lib.jira.utils import (
     convert_to_jira_datetime_string, find_allowed_values, make_fields_lookup
 )
 
+LAST_UPDATED_AT = 'Github PR Last Updated At'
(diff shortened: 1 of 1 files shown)


openedx_webhooks/jira/tasks.py
```diff
@@ -5,11 +5,27 @@
...
+    'LATEST_ACTION',
+    'LATEST_BY_EDX',
+])
+
+GH_PR_ACTIVITY = GithubActivity(
+    LAST_UPDATED_AT='Github PR Last Updated At',
+    LAST_UPDATED_BY='Github PR Last Updated By',
+    LATEST_ACTION='Github Latest Action',
+    LATEST_BY_EDX='Github Latest Action by edX',
+)
```
nedbat: This seems like a long way to go when four constants would have sufficed, but ok.

nedbat:
:+1:

PR #94, opened 2016-11-09
Handle robot and unknown person PR activities

gsong:
1. Ignore incoming events from known bots (entries with `is_robot: true` in `people.yaml`).
2. For users not in `people.yaml` assume they are not edX users but still process the event.

/cc @edx/ospr


nedbat:
:+1:

PR #95, opened 2016-11-10
Fix: Test for known user before robot

gsong:
We were testing for a robot before an unknown user, which would cause an exception if the user is unknown.

* Test for an unknown user first
* Add more tests!
* Various cleanup

/cc @edx/ospr


openedx_webhooks/github/dispatcher/rules/github_activity.py
```diff
@@ -38,7 +38,8 @@ def _is_sender_edx_user(event):
         bool
     """
     try:
-        return event.sender.is_edx_user
+        event.sender
+        return True
```
nedbat: This seems like a strange way to do it: depending on event.sender to raise NotFoundError?  As a property, it feels more natural for it to return None in this case.

openedx_webhooks/github/dispatcher/rules/test/test_github_activity.py
```diff
@@ -0,0 +1,88 @@
...
+    ):
+        payload = _payload.copy()
+        payload['sender']['login'] = login
+        assert (
+            _process(github_client, jira_client, 'issue_comment', payload)
+            is None
+        )
+        (
+            openedx_webhooks.github.dispatcher.rules.github_activity
+            ._find_issues_for_pull_request.assert_called_once_with(
```
nedbat: This code would read better if you assign some intermediate values.

openedx_webhooks/github/dispatcher/rules/test/test_github_activity.py
```diff
@@ -0,0 +1,88 @@
...
+
+class TestProcess:
+    def _test_user(
+            self, github_client, jira_client, _payload, login, is_edx_user
+    ):
+        payload = _payload.copy()
+        payload['sender']['login'] = login
+        assert (
+            _process(github_client, jira_client, 'issue_comment', payload)
+            is None
```
nedbat: `_process` has no return at all, it seems odd to assert it returns None.

nedbat:
:+1:

PR #100, opened 2017-10-20
Provide a version for tini

nedbat:


nedbat:
@gsong this fixed this error:
```
standard_init_linux.go:185: exec user process caused "exec format error"
```

PR #101, opened 2017-10-25
WIP: Rewrite handling of new GitHub PR

gsong:
Convert new GitHub PR into OSPR if appropriate. Also make the right comments on the PR itself as needed.

PR #102, opened 2017-10-27
WIP: Rewrite handling of new GitHub PR

gsong:
Convert new GitHub PR into OSPR if appropriate. Also make the right comments on the PR itself as needed.

Replaces #101.


openedx_webhooks/github/dispatcher/actions/new_pull_request_helpers/ospr_github.py
```diff
@@ -0,0 +1,57 @@
...
+)
+
+CONTRIBUTING_MSG = (
+    "Please see the [CONTRIBUTING]({}) file for more information."
+).format(CONTRIBUTING_URL)
+
+
+@inject_gh
+def create_ospr_comment(gh, event, ospr_info):
+    # TODO:
```
gsong: See the current Jinja template for when to conditionally include these messages.

openedx_webhooks/github/dispatcher/actions/new_pull_request_helpers/ospr_jira.py
```diff
@@ -0,0 +1,12 @@
...
+from __future__ import (
+    absolute_import, division, print_function, unicode_literals
+)
+
+from .....lib.jira.decorators import inject_jira
+
+
+@inject_jira
+def create_ospr(jira, event, workflow_stage='Needs Triage'):
+    # TODO: implement
```
gsong: Obviously this still needs to be implemented, it should return `{'label': 'OSPR-1234', 'url': 'https://xyz.com'}`.

The idea of `workflow_stage` is to be able to put the OSPR directly into a specific workflow stage. For example, if edX requires CLA, then it can go directly to the "Awaiting Author" stage, instead of someone having to triage.

gsong: @nedbat The only remaining thing is to create the actual OSPR in JIRA:

1. This should probably be moved to the `openedx_webhooks.jira.tasks` module, since it has to do with JIRA, not GitHub.
2. You can see [`update_latest_github_activity`](https://github.com/edx/openedx-webhooks/blob/development/openedx_webhooks/jira/tasks.py#L48) for clues on how this function should be implemented.

PR #104, opened 2018-05-15
adding check for former edX employee

mduboseedx:
This should check the before entry in people.yaml to see if the PR was created while a former employee was at edX. 

Do we have data to test this in staging?

### DIFF
--- a/openedx_webhooks/info.py
+++ b/openedx_webhooks/info.py
@@ -48,6 +48,23 @@ def get_orgs(key):
     orgs = get_orgs_file()
     return set(o for o, info in orgs.items() if info.get(key, False))
 
+def get_person_certain_time(person, certain_time):
+    """
+    Return person data structure for a particular time
+
+    Arguments:
+        person: dict of a Github user info from people.yaml in repo-tools-data
+        certain_time: datetime.datetime object used to determine the state of the person
+
+    """
+    for before_date in sorted(person.get("before", {})):
+        if certain_time.date() <= before_date:
+            before_person = person["before"][before_date]
+            update_person = person.copy()
+            update_person.update(before_person)
+            return update_person
+    return person
+
 
 def is_internal_pull_request(pull_request):
     """
@@ -77,7 +94,6 @@ def is_bot_pull_request(pull_request):
     """
     return pull_request["user"]["type"] == "Bot"
 
-
 def _is_pull_request(pull_request, kind):
     """
     Is this pull request of a certain kind?
@@ -101,6 +117,8 @@ def _is_pull_request(pull_request, kind):
     if person.get("expires_on", date.max) <= created_at.date():
         # This person's agreement has expired.
         return False
+
+    person = get_person_certain_time(people[author], created_at)
 
     if person.get(kind, False):
         # This person has the flag personally.
--- a/tests/test_info.py
+++ b/tests/test_info.py
@@ -6,7 +6,7 @@ from datetime import datetime
 import pytest
 
 import openedx_webhooks.info
-from openedx_webhooks.info import get_orgs, is_internal_pull_request
+from openedx_webhooks.info import get_orgs, get_people_file, is_internal_pull_request, get_person_certain_time
 
 pytestmark = pytest.mark.usefixtures('mock_github')
 
@@ -42,7 +42,7 @@ def test_ex_edx_employee():
 
 def test_ex_edx_employee_old_pr():
     created_at = datetime(2014, 1, 1)
-    pr = make_pull_request("mmprandom", created_at=created_at)
+    pr = make_pull_request("jarv", created_at=created_at)
     assert is_internal_pull_request(pr)
 
 def test_never_heard_of_you():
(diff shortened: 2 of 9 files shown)


openedx_webhooks/info.py
```diff
@@ -111,4 +118,11 @@ def _is_pull_request(pull_request, kind):
         # This person's institution has the flag.
         return True
 
+    if person.get("before"):
+        for before_date in sorted(person.get("before").keys()):
+          if created_at.date() <= before_date:
+            if person.get("before")[before_date].get("institution") in the_orgs:
+              # This person was an employee of edX when this PR was created, but has since left
+              return True
+
```
nedbat: We talked yesterday about moving this logic.  It would be good to make a new function that takes a person and a date, and returns the person data structure as of that date.  That is a good unit to be a function all its own, and then it can be separately tested.

openedx_webhooks/info.py
```diff
@@ -48,6 +48,18 @@ def get_orgs(key):
     orgs = get_orgs_file()
     return set(o for o, info in orgs.items() if info.get(key, False))
 
+def get_person_time_of_pull_request(person, created_at):
```
nedbat: In isolation, this function has nothing to do with pull requests.  It takes a person and a date, and gives you the state of the person at that date.  We happen to use it with a pull request created_at date, but there's no reason to include that fact in the function itself.

We can give this a more generic name, and the created_at parameter can be given a more generic name.

openedx_webhooks/info.py
```diff
@@ -48,6 +48,18 @@ def get_orgs(key):
...
+    """
+    Return person data structure as of time of pull request
```
nedbat: Let's state the types of the arguments.  I'm not sure what `created_at` needs to be.

openedx_webhooks/info.py
```diff
@@ -48,6 +48,18 @@ def get_orgs(key):
...
+    """
+    for before_date in sorted(person.get("before", {})):
+        if created_at.date() <= before_date:
```
nedbat: I'm curious why you needed `.date()` here.

openedx_webhooks/info.py
```diff
@@ -48,6 +48,18 @@ def get_orgs(key):
...
+            before_person = person.get("before")[before_date]
```
nedbat: `.get` here can just be a regular key lookup, since we know that there's a "before" key.

openedx_webhooks/info.py
```diff
@@ -48,6 +48,18 @@ def get_orgs(key):
...
+            update_person = person.copy()
+            update_person.update(before_person)
```
nedbat: I'd need to look at the data again to be sure about this being enough.  If the current person has extra values that aren't mentioned in the before data, then the updated person will have a mix of old and new.  I'm not sure if that happens, or if it matters?

openedx_webhooks/info.py
```diff
@@ -48,6 +48,18 @@ def get_orgs(key):
...
+            return update_person
+    return person
```
nedbat: We need some unit tests for this bad boy. :)

openedx_webhooks/info.py
```diff
@@ -103,8 +107,8 @@ def _is_pull_request(pull_request, kind):
         # We don't know this person!
         return False
 
-    person = people[author]
     created_at = parse_date(pull_request["created_at"]).replace(tzinfo=None)
+    person = get_person_time_of_pull_request(people[author], created_at)
     if person.get("expires_on", date.max) <= created_at.date():
```
nedbat: This is another twisty thing: is it OK to check "expires_on" in the past-person? Or should we check it on the current person?

mduboseedx: The "expires_on" key is not within any of the "before" dictionaries so if a user has "expire_on" set, then that PR will be seen as an OSPR regardless if the user info is updated with the "before" information. Should this ordering still be changed to be make this cleaner?

openedx_webhooks/info.py
mduboseedx: The person will have a mix of old and new values, but I don't believe that actually matters. We are only checking the "agreement", "institution" and "expires_on" keys as the other information, other than the Github username, is currently irrelevant (I think).

openedx_webhooks/info.py
```diff
@@ -48,13 +48,18 @@ def get_orgs(key):
     orgs = get_orgs_file()
     return set(o for o, info in orgs.items() if info.get(key, False))
 
-def get_person_time_of_pull_request(person, created_at):
+def get_updated_person(person, updated_time):
```
nedbat: I'm not sure how finely detailed to get here.  Ideally the name of the function would include the idea of "date in the past" or something. "Updated" is pretty vague, though the docstring covers it, so....

openedx_webhooks/info.py
nedbat: I think if "expires_on" can never be changed by "get_updated_person", then we should check it before we call get_updated_person.

tests/test_info.py
```diff
@@ -6,10 +6,21 @@
 import pytest
 
 import openedx_webhooks.info
-from openedx_webhooks.info import get_orgs, is_internal_pull_request
+from openedx_webhooks.info import get_orgs, is_internal_pull_request, get_updated_person
+
+import os
```
nedbat: Imports from the stdlib should go in a section at the top.  1) stdlib, 2) third-party modules, 3) our own modules.

mduboseedx:
I've add new test data and changed the tests, but I'm not sure if they are actually useful tests.

tests/test_info.py
```diff
@@ -78,6 +79,7 @@ def test_updated_person_has_institution():
     created_at = datetime(2014, 1, 1)
     updated_person = get_person_certain_time(people["jarv"], created_at)
     assert "institution" in updated_person
+    assert updated_person["institution"] == "edX"
```
nedbat: This line makes line 81 unnecessary.

PR #106, opened 2018-07-19
Removing AUTHORS line from PR comments

mduboseedx:
I simply removed the AUTHORS line from the template and then changed the tests to check that it didn't appear in any comments. The tests pass, but is that all that needs to be done?

### DIFF
--- a/openedx_webhooks/templates/github_community_pr_comment.md.j2
+++ b/openedx_webhooks/templates/github_community_pr_comment.md.j2
@@ -26,10 +26,6 @@ GitHub pull request interface. As a reminder,
     [signed contributor agreement](http://open.edx.org/sites/default/files/wysiwyg/individual-contributor-agreement.pdf)
     or indicated your institutional affiliation.
   {%- endif %}
-  {%- if not in_authors_file %}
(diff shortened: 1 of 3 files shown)


openedx_webhooks/templates/github_community_pr_comment.md.j2
```diff
@@ -26,10 +26,6 @@ GitHub pull request interface. As a reminder,
     [signed contributor agreement](http://open.edx.org/sites/default/files/wysiwyg/individual-contributor-agreement.pdf)
     or indicated your institutional affiliation.
   {%- endif %}
-  {%- if not in_authors_file %}
```
nedbat: Let's get rid of `in_authors_file` completely.  We don't need to calculate it in the view any more, and then we don't need it in the next clause here in the template, and then we can simplify the template.

mduboseedx:
@nedbat I made some changes and I think I've removed `in_authors_file` from the other areas of the code.

openedx_webhooks/templates/github_community_pr_comment.md.j2
nedbat: We now have two consecutive clauses with "if not has_signed_agreement", so we can combine them.

mduboseedx:
@nedbat I think his has now addressed your comments

PR #107, opened 2018-09-04
adding cherry pick comment to internal PRs

mduboseedx:
The tests pass, but I think I need to add a new test when specifying the TEST_OPENEDX_RELEASE_NAME environment variable. I'm not quite sure how to do that properly.


openedx_webhooks/tasks/github.py
```diff
@@ -28,6 +29,9 @@
 
 COVERLETTER_MARKER = "<!-- open edx coverletter -->"
 
+open_edx_release = os.environ.get('OPENEDX_RELEASE_NAME')
+test_open_edx_release = os.environ.get('TEST_OPENEDX_RELEASE_NAME')
```
nedbat: Make this be a list if GitHub nicks who should get the test message.

openedx_webhooks/tasks/github.py
```diff
@@ -85,6 +89,14 @@ def pull_request_opened(self, pull_request, ignore_internal=True, check_contract
         # not an open source pull request, don't create an issue for it
         msg = "@{user} opened PR #{num} against {repo} (internal PR)".format(user=user, repo=repo, num=num)
         log_info(self.request, msg)
+        # new release candidate for Open edX is available, ask internal PR if should be cherry picked
+        if open_edx_release:
+            github_prepare_cherry_pick_comment(github, pr, open_edx_release)
+            return None, True
+        elif test_open_edx_release:
+            if user in ['mduboseedx']:
+                github_prepare_cherry_pick_comment(github, pr, open_edx_release)
```
nedbat: This line is the same as 94.  Two things about that: 1) `open_edx_release` will be None here, so it won't work, and 2) ideally the action line will occur only once, conditionalized on a `doit` boolean that is computed based on the earlier conditions.  I suppose putting these two ideas together, we could keep the action lines separate so they could have different third arguments?

openedx_webhooks/tasks/github.py
```diff
@@ -357,6 +369,25 @@ def get_jira_issue_key(pull_request):
     return None
 
 
+def github_prepare_cherry_pick_comment(github, pull_request, open_edx_release):
+    """
+    Prepares a cherry pick comment used for internal engineers during the window between
+    an Open edX release candidate and the official release.
+
+    * fun_facts includes a piece of trivia in an effort to get an engineer's attention
```
nedbat: I'm not sure why fun_facts is mentioned in this docstring?

nedbat: Maybe this should be a comment in the body instead?

openedx_webhooks/tasks/github.py
```diff
@@ -357,6 +369,25 @@ def get_jira_issue_key(pull_request):
...
+    """
+    repos = get_repos_file()
+    if repo in repos:
```
nedbat: `repo` seems to be undefined.  Do we need this test here?  Will this function be called for repos we shouldn't handle?

openedx_webhooks/tasks/github.py
```diff
@@ -357,6 +369,25 @@ def get_jira_issue_key(pull_request):
     return None
 
 
+def github_prepare_cherry_pick_comment(github, pull_request, open_edx_release):
+    """
+    Prepares a cherry pick comment used for internal engineers during the window between
+    an Open edX release candidate and the official release.
```
nedbat: This docstring isn't quite right: this function posts a comment.

openedx_webhooks/tasks/github.py
```diff
@@ -399,6 +430,20 @@ def github_contractor_pr_comment(pull_request):
         number=pull_request["number"],
     )
 
+def github_internal_cherrypick_comment(pull_request, release_name, fun_fact_question, fun_fact_answer):
+    """
+    During the window between a release candidate and the official release of Open edX,
+    ask Internal authors if their PRs should be cherry picked onto the release candidate.
```
nedbat: This docstring isn't quite right: this function formats a comment.

openedx_webhooks/templates/github_internal_cherrypick_comment.md.j2
```diff
@@ -0,0 +1,13 @@
+{% filter replace("\n", " ")|trim %}
+Well done, @{{ user }}! We have recently cut the {{ open_edx_release_name }} release candidate
```
nedbat: "release candidate" isn;t the right term here.  Maybe, "We've recently created the Hawthorn master branches"

openedx_webhooks/templates/github_internal_cherrypick_comment.md.j2
```diff
@@ -0,0 +1,13 @@
...
+and perhaps this PR should be cherry picked onto the branch. While you consider cherry
+picking, please enjoy this Fun fact:
```
nedbat: Either Fun should be lower-case, or fact should be capitalized :)

tests/test_pr_comments.py
```diff
@@ -215,6 +220,18 @@ def test_has_contractor_comment_no_comments(app, reqctx, requests_mocker):
     assert result is False
 
 
+def test_has_internal_cherry_pick_comment_release_environ_set(app):
+    pr = make_pull_request(user='nedbat')
+    open_edx_release = "Matt's Open edX Release"
+    fun_facts = get_fun_fact_file()
+    question, answer = random.choice(list(fun_facts.items()))
```
nedbat: This is doing too much work for this test, and duplicates lines in the product code.  You can test `github_internal_cherrypick_comment` by passing it literal strings and checking the output.

tests/test_pr_comments.py
```diff
@@ -215,6 +220,18 @@ def test_has_contractor_comment_no_comments(app, reqctx, requests_mocker):
...
+    with app.test_request_context('/'):
+        comment = github_internal_cherrypick_comment(pr, open_edx_release, question, answer)
+    assert "Matt's Open edX Release" in comment
+    assert "Which animal's eye" in comment
+    assert "hide their head in the sand" in comment
+
+
```
nedbat: You need a test of the logic in `pull_request_opened`, and in `github_prepare_cherry_pick_comment`.

openedx_webhooks/tasks/github.py
```diff
@@ -369,23 +378,22 @@ def get_jira_issue_key(pull_request):
...
-        comment_resp = github.post(url, json=comment)
-        log_request_response(self.request, comment_resp)
-        comment_resp.raise_for_status()
+    # fun_facts is a dictionary of trivia questions and answers
+    fun_facts = get_fun_fact_file()
+    question, answer = random.choice(list(fun_facts.items()))
+    comment = {
+        "body": github_internal_cherrypick_comment(pull_request, open_edx_release, question, answer),
+    }
+    url = "/repos/{repo}/issues/{num}/comments".format(repo=pull_request["base"]["repo"]["full_name"], num=pull_request["number"])
```
nedbat: This line would be easier to read on multiple lines:
```
url = "/repos/{repo}/issues/{num}/comments".format(
    repo=pull_request["base"]["repo"]["full_name"], 
    num=pull_request["number"],
)
```

tests/test_pr_comments.py
```diff
@@ -220,16 +221,66 @@ def test_has_contractor_comment_no_comments(app, reqctx, requests_mocker):
...
-    question, answer = random.choice(list(fun_facts.items()))
+    question = "This is a question"
+    answer = "This is an answer"
     with app.test_request_context('/'):
         comment = github_internal_cherrypick_comment(pr, open_edx_release, question, answer)
     assert "Matt's Open edX Release" in comment
-    assert "Which animal's eye" in comment
-    assert "hide their head in the sand" in comment
+    assert "This is a question" in comment
+    assert "This is an answer" in comment
```
nedbat: We should probably also assert about some other contents of the message.

tests/test_pr_comments.py
```diff
@@ -215,6 +221,73 @@ def test_has_contractor_comment_no_comments(app, reqctx, requests_mocker):
...
+        pr = make_pull_request(user='nedbat')
+        requests_mocker.post(
+            "https://api.github.com/repos/edx/edx-platform/issues/1/comments",
+            headers={"Content-Type": "application/json"},
+        )
+        with reqctx:
+            app.preprocess_request()
+            result = openedx_webhooks.tasks.github.pull_request_opened(pr)
+        assert result[1] is False
+        assert requests_mocker.last_request.text is None
```
mduboseedx: I did not use `requests_mocker.called` since that should evaluate to True as there are other calls to the people.yaml and org.yaml files

tests/test_pr_comments.py
```diff
@@ -215,6 +221,76 @@ def test_has_contractor_comment_no_comments(app, reqctx, requests_mocker):
...
+        pr = make_pull_request(user='nedbat')
+        requests_mocker.post(
+            "https://api.github.com/repos/edx/edx-platform/issues/1/comments",
+            headers={"Content-Type": "application/json"},
+        )
+        with reqctx:
+           app.preprocess_request()
+           result = openedx_webhooks.tasks.github.pull_request_opened(pr)
+        assert result[1] is True
+        assert "Test Release" in requests_mocker.last_request.json()['body']
```
nedbat: We should assert that the last request is the one we're interested in, by looking at the url.

tests/test_pr_comments.py
```diff
@@ -215,6 +221,76 @@ def test_has_contractor_comment_no_comments(app, reqctx, requests_mocker):
...
+def test_open_edx_dry_run_environment_var_set(app, reqctx, requests_mocker):
+    test_open_edx_release = 'mduboseedx,nedbat,fakeuser'
+    with mock.patch.dict('os.environ', {'GITHUB_USERS_CHERRY_PICK_MESSAGE_TEST': test_open_edx_release}):
+        pr = make_pull_request(user='nedbat')
+        requests_mocker.post(
+            "https://api.github.com/repos/edx/edx-platform/issues/1/comments",
+            headers={"Content-Type": "application/json"},
+        )
+        with reqctx:
+           app.preprocess_request()
```
nedbat: 3-space indent!?

tests/test_pr_comments.py
```diff
@@ -215,6 +221,76 @@ def test_has_contractor_comment_no_comments(app, reqctx, requests_mocker):
...
+        pr = make_pull_request(user='nedbat')
+        requests_mocker.post(
+            "https://api.github.com/repos/edx/edx-platform/issues/1/comments",
+            headers={"Content-Type": "application/json"},
+        )
+        with reqctx:
+            app.preprocess_request()
+            result = openedx_webhooks.tasks.github.pull_request_opened(pr)
+        assert result[1] is True
+        assert "Open edX Release" in requests_mocker.last_request.json()['body']
```
nedbat: Ditto for asserting on the url.

PR #108, opened 2018-10-30
added make upgraded target and updated dependencies

mduboseedx:
I also ran the make upgrade which updated a number of dependencies.


nedbat:
@mduboseedx As we discussed, unpin all the flask stuff, and let's try it!

PR #109, opened 2018-11-09
WIP on checking committers CLA on PR

mduboseedx:


tests/test_pr_comments.py
```diff
@@ -123,6 +130,65 @@ def test_community_pr_comment_no_author(app):
...
+    committers = make_committers(["phony", "notreal"])
+    with app.test_request_context('/'):
+        comment = github_community_pr_comment(pr, jira, people, committers)
+    assert "[FOO-1](https://openedx.atlassian.net/browse/FOO-1)" in comment
+    assert "can't start reviewing your pull request" in comment
+    assert "commit authors will also need to submit a signed contributor" in comment
+    assert "phony" in comment
+    assert "notreal" in comment
+
+def test_community_pr_committers_api_called(app, reqctx, requests_mocker):
```
mduboseedx: This test does not pass at the moment

openedx_webhooks/tasks/github.py
```diff
@@ -413,17 +416,27 @@ def github_community_pr_comment(pull_request, jira_issue, people=None):
     people = {user.lower(): values for user, values in people.items()}
     pr_author = pull_request["user"]["login"].decode('utf-8').lower()
     created_at = parse_date(pull_request["created_at"]).replace(tzinfo=None)
+    missing_committers = []
+    if commits:
```
nedbat: Is this to check for empty commits, or commits == None? If it's only for empty, then you can just let the for loop take care of it.

openedx_webhooks/tasks/github.py
```diff
@@ -413,17 +416,27 @@ def github_community_pr_comment(pull_request, jira_issue, people=None):
...
+        for committer in commits:
+            try:
+                if commits[committer]["author"]["login"] not in people:
+                    missing_committers.append(commits[committer]["author"]["login"])
+            except TypeError:
```
nedbat: What TypeError is this?

openedx_webhooks/tasks/github.py
```diff
@@ -413,17 +416,27 @@ def github_community_pr_comment(pull_request, jira_issue, people=None):
     people = {user.lower(): values for user, values in people.items()}
     pr_author = pull_request["user"]["login"].decode('utf-8').lower()
     created_at = parse_date(pull_request["created_at"]).replace(tzinfo=None)
+    missing_committers = []
```
nedbat: This should probably be a set, so that three commits by the same newb doesn't mention them three times in the comment.

openedx_webhooks/tasks/github.py
```diff
@@ -413,17 +416,27 @@ def github_community_pr_comment(pull_request, jira_issue, people=None):
...
+    if commits:
+        for committer in commits:
+            try:
+                if commits[committer]["author"]["login"] not in people:
+                    missing_committers.append(commits[committer]["author"]["login"])
```
nedbat: These two lines use the same complex expression: pull it into its own variable.

openedx_webhooks/tasks/github.py
```diff
@@ -413,17 +416,27 @@ def github_community_pr_comment(pull_request, jira_issue, people=None):
...
+            except TypeError:
+                pass
     # does the user have a valid, signed contributor agreement?
     has_signed_agreement = (
         pr_author in people and
-        people[pr_author].get("expires_on", date.max) > created_at.date()
+        people[pr_author].get("expires_on", date.max) > created_at.date() and not
+        missing_committers
```
nedbat: Looks like we consider the expires_on for the pr_author, but not for the commiters?  I smell a function in our future...

openedx_webhooks/templates/github_community_pr_comment.md.j2
```diff
@@ -22,9 +22,16 @@ GitHub pull request interface. As a reminder,
 
 {% filter replace("\n", " ")|trim %}
   {%- if not has_signed_agreement %}
-    We can't start reviewing your pull request until you've submitted a
+    We can't start reviewing your pull request until you and the other commit authors submit a
     [signed contributor agreement](http://open.edx.org/sites/default/files/wysiwyg/individual-contributor-agreement.pdf)
-    or indicated your institutional affiliation. Please see the [CONTRIBUTING](https://github.com/edx/edx-platform/blob/master/CONTRIBUTING.rst)
+    or indicated your institutional affiliation.
+    {%- if missing_committers %}
+    The following commit authors will also need to submit a signed contributor agreement:
```
nedbat: Use indentation.

openedx_webhooks/templates/github_community_pr_comment.md.j2
```diff
@@ -22,9 +22,16 @@ GitHub pull request interface. As a reminder,
 
 {% filter replace("\n", " ")|trim %}
   {%- if not has_signed_agreement %}
-    We can't start reviewing your pull request until you've submitted a
+    We can't start reviewing your pull request until you and the other commit authors submit a
```
nedbat: I wonder if we need to make this sentence more sophisticated, so that it reads properly in all cases.

tests/test_pr_comments.py
```diff
@@ -123,6 +130,65 @@ def test_community_pr_comment_no_author(app):
...
+def test_community_pr_comment_no_author_cla_has_committers_cla(app):
+    pr = make_pull_request(user="FakeUser")
+    jira = make_jira_issue(key="FOO-1")
+    people = get_people_file()
+    committers = make_committers(["mduboseedx", "nedbat"])
+    with app.test_request_context('/'):
+        comment = github_community_pr_comment(pr, jira, people, committers)
+    assert "[FOO-1](https://openedx.atlassian.net/browse/FOO-1)" in comment
+    assert "can't start reviewing your pull request" in comment
+    assert "commit authors will also need to submit a signed contributor" not in comment
```
nedbat: This is where the sentence seems wrong: all these commit authors have a CLA, but it says "commit authors will also need to submit".

tests/test_pr_comments.py
nedbat: Why not?

PR #110, opened 2018-11-19
pinning Flask-dance and upgrading

mduboseedx:


nedbat:
Ideally, we'd try to upgrade our code to use the latest versions.  Let's make a ticket to re-examine FlaskDance, and then merge this.

PR #111, opened 2018-12-05
rescan all repos now uses the GH api

mduboseedx:
This is undoubtedly a work in progress. I haven't added any new tests since nothing in the openedx_webhooks/github_views.py file has actually been tested. Do you think a new file in the test directory is needed?

### DIFF
--- a/openedx_webhooks/github_views.py
+++ b/openedx_webhooks/github_views.py
@@ -153,11 +153,13 @@ def rescan():
     if inline:
         return jsonify(rescan_repository(repo))
 
-    if repo == 'all':
-        repos = get_repos_file().keys()
+    if repo.startswith('all:'):
+        org = repo[4:]
+        org_url = "https://api.github.com/orgs/{org}/repos".format(org=org)
+        repo_names = [repo_name['full_name'] for repo_name in paginated_get(org_url)]
         workflow = group(
-            rescan_repository.s(repo, wsgi_environ=minimal_wsgi_environ())
-            for repo in repos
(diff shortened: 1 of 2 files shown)


openedx_webhooks/github_views.py
```diff
@@ -154,7 +154,11 @@ def rescan():
         return jsonify(rescan_repository(repo))
 
     if repo == 'all':
-        repos = get_repos_file().keys()
+        org_url = "https://api.github.com/orgs/edx/repos"
```
nedbat: We'll need to deal with other orgs also somehow.  That was one thing repos.yaml handled for us, since it has a few edx-solutions repos also.

openedx_webhooks/github_views.py
```diff
@@ -154,7 +154,11 @@ def rescan():
...
+        edx_repos = paginated_get(org_url)
+        repos = []
+        for edx_repo in edx_repos:
+            repos.append(edx_repo['full_name'])
```
nedbat: These four lines can be a list comprehension:
```
repos = [repo['full_name'] for repo in paginated_get(org_url)]
```

nedbat: Though it would be clearer to call this "repo_names" instead of "repos"

mduboseedx:
I tested this by hitting the button in Staging and using "all:openedx". I didn't see an error in the logs, but instead had this:

INFO: Created 0 JIRA issues on repo openedx/openedx-conference-website. PRs are [] [openedx_webhooks.tasks]

Do we give this a try in prod?

PR #118, opened 2018-12-12
adding ability to skip jira creation for specific orgs

mduboseedx:
Surprise surprise, some of the tests are not working properly. Let's discuss!


mduboseedx:
Tests are passing now and the new code is covered.

I'll need to add the no_jira_org.yaml file to repo-tools-data before this should be merged.

openedx_webhooks/info.py
```diff
@@ -97,6 +101,19 @@ def is_bot_pull_request(pull_request):
...
     return pull_request["user"]["type"] == "Bot"
 
+def is_no_jira_org_pull_request(pull_request):
+    """
+    Was this pull request created within a GitHub organization that should not have a
+    Jira ticket created?
+    """
+    orgs = get_no_jira_orgs_file()
+    for org in orgs:
+        match = re.search(org + "/", pull_request["base"]["repo"]["full_name"])
```
nedbat: This is more easily done with `startswith(org + "/")`

openedx_webhooks/info.py
```diff
@@ -97,6 +101,19 @@ def is_bot_pull_request(pull_request):
...
+        if match:
+            if orgs[org]['Jira'] == False:
```
nedbat: `if not orgs[org]['Jira']:`

openedx_webhooks/info.py
```diff
@@ -97,6 +101,19 @@ def is_bot_pull_request(pull_request):
     """
     return pull_request["user"]["type"] == "Bot"
 
+def is_no_jira_org_pull_request(pull_request):
```
nedbat: I think it's easier to reason about things expressed positively. This could be `should_add_jira_for_pull_request`, or `is_jira_pull_request` (No need for org in the name, since the caller doesn't need to know what info is being used to decide to do jira or not.)

openedx_webhooks/tasks/github.py
```diff
@@ -75,6 +76,15 @@ def pull_request_opened(self, pull_request, ignore_internal=True, check_contract
         # Bots never need OSPR attention.
         return None, False
 
+    if is_no_jira_org_pull_request(pr):
```
nedbat: The JIRA/no-JIRA decision should be lighter-weight.  Continue with the main flow of the function.

nedbat: Call `github_community_pr_comment` eventually just as before, but with the JIRA argument set to None.

tests/test_pr_comments.py
```diff
@@ -60,6 +61,15 @@ def make_jira_issue(key="ABC-123"):
     'failing due to '
     '`BetamaxError: A request was made that could not be handled.`'
 ))
+
+def make_request_mocker_history(requests_mocker, url, method):
+    needed_request = None
+    history = requests_mocker.request_history
+    for request_mock in history:
+        if request_mock.url == url and request_mock.method == method:
+            needed_request = request_mock
```
nedbat: You could just return from this line.

tests/test_pr_comments.py
```diff
@@ -60,6 +61,15 @@ def make_jira_issue(key="ABC-123"):
     'failing due to '
     '`BetamaxError: A request was made that could not be handled.`'
 ))
+
+def make_request_mocker_history(requests_mocker, url, method):
```
nedbat: I'm not sure "make" is the right verb here.  This feels like "find_mocked_url"

nedbat: It'll be easier to use this if you default method to GET.

PR #119, opened 2019-05-21
Add HTTP basic auth to homepage.

doctoryes:
Based on:
http://flask.pocoo.org/snippets/8/


doctoryes:
@nedbat  FYI

nedbat:
It's live and it works! Thanks! :) <3

PR #144, opened 2021-07-02
feat: Add branches field for core committers

sarina:
Add a new, optional field `branches` that gives core committers access to all branches that are specified in the set.

`branches` is a set of one or more specified branches that CCs have merge rights to.

`branches` can be a specific branch, such as `open-release/maple.1`, or a wildcard such as `open-release*`

This is dependent on https://github.com/edx/repo-tools-data-schema/pull/3

### DIFF
--- a/openedx_webhooks/info.py
+++ b/openedx_webhooks/info.py
@@ -142,7 +142,8 @@ def _is_pull_request(pull_request: PrDict, kind: str) -> bool:
 
 def is_committer_pull_request(pull_request: PrDict) -> bool:
     """
-    Was this pull request created by a core committer for this repo?
+    Was this pull request created by a core committer for this repo
+    or branch?
     """
     person = _pr_author_data(pull_request)
     if person is None:
@@ -152,6 +153,7 @@ def is_committer_pull_request(pull_request: PrDict) -> bool:
 
     repo = pull_request["base"]["repo"]["full_name"]
     org = repo.partition("/")[0]
+    branch = pull_request["base"]["ref"]
     commit_rights = person["committer"]
     if not commit_rights:
         return False
@@ -161,6 +163,12 @@ def is_committer_pull_request(pull_request: PrDict) -> bool:
     if "repos" in commit_rights:
         if repo in commit_rights["repos"]:
             return True
+    if "branches" in commit_rights:
+        for access_branch in commit_rights["branches"]:
+            if access_branch.endswith("*") and branch.startswith(access_branch[:-1]):
+                return True
+            elif branch == access_branch:
+                return True
     return False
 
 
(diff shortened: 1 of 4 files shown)


sarina:
The failing test is because the validation in the data schema fails

openedx_webhooks/info.py
```diff
@@ -152,6 +153,9 @@ def is_committer_pull_request(pull_request: PrDict) -> bool:
 
     repo = pull_request["base"]["repo"]["full_name"]
     org = repo.partition("/")[0]
+    # Check if this is a named release branch
+    # compare_url is "https://api.github.com/repos/octocat/Hello-World/compare/{base}...{head}"
+    open_release = "open-release" in pull_request["base"]["compare_url"].split("...")[0]
```
nedbat, line 158: Instead of a boolean "open-release" key, why not a "branch" key with a pattern for a value?

sarina: I just edited the description to say that was a deliberate choice because I think it's easier and more readable, but I'm open to disagreements.

sarina: (particularly, this matches the bot's behavior which is hardcoded)

nedbat, line 158: `pull_request["base"]["ref"]` will be a more direct check. its value is `"open-release/lilac.master"`.

nedbat: It feels to me like it's not much more to use a pattern here, and will perhaps be more understandable in people.yaml, since "branch: open-release/*" is pretty clear in Git terms, but "open-release: true" is less so.

sarina:
@nedbat I chose to do `base_branches` instead of `branches`, but I realize I could probably be fancier. IDK i'm just doing this for fun. Let me know your thoughts.

openedx_webhooks/info.py
```diff
@@ -152,6 +153,7 @@ def is_committer_pull_request(pull_request: PrDict) -> bool:
 
     repo = pull_request["base"]["repo"]["full_name"]
     org = repo.partition("/")[0]
+    base_branch = pull_request["base"]["ref"].split('/')[0]
```
nedbat, line 156: I guess I would still prefer a pattern to match, than to hard-code the idea that the part before the slash is significant.

sarina: FINE lol so sorry I'm being lazy about this :) try now

openedx_webhooks/info.py
```diff
@@ -161,6 +163,12 @@ def is_committer_pull_request(pull_request: PrDict) -> bool:
     if "repos" in commit_rights:
         if repo in commit_rights["repos"]:
             return True
+    if "branches" in commit_rights:
+        for access_branch in commit_rights["branches"]:
+            if access_branch.endswith("*") and branch.startswith(access_branch[:-1]):
```
nedbat, line 168: OK! The fnmatch module is useful for these kinds of patterns, but this will work.

nedbat (approved):

PR #145, opened 2021-07-28
feat: add Github label to PR if CLA still needed

stvstnfrd:
This is part of a larger effort to ensure CLAs are signed pre-merge,
when needed.

It serves as a minimal first step, as well as providing an opportunity
for me (stvstnfrd) to familiarize myself with the Heroku workflow.

### DIFF
--- a/openedx_webhooks/tasks/pr_tracking.py
+++ b/openedx_webhooks/tasks/pr_tracking.py
@@ -263,6 +263,7 @@ def desired_support_state(pr: PrDict) -> Optional[PrDesiredInfo]:
     if not has_signed_agreement:
         desired.bot_comments.add(BotComment.NEED_CLA)
         desired.jira_initial_status = "Community Manager Review"
+        desired.github_labels.add('NEED-CLA')
 
     if state == "closed":
         desired.jira_status = "Rejected"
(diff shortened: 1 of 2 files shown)


openedx_webhooks/tasks/pr_tracking.py
```diff
@@ -263,6 +263,7 @@ def desired_support_state(pr: PrDict) -> Optional[PrDesiredInfo]:
     if not has_signed_agreement:
         desired.bot_comments.add(BotComment.NEED_CLA)
         desired.jira_initial_status = "Community Manager Review"
+        desired.github_labels.add('NEED-CLA')
```
stvstnfrd, line 266: Cribbed from above (in this method).

stvstnfrd, line 266: Naming convention is fairly arbitrary. I went with:
- ALL CAPS: to make it stand out
- hyphen instead of underscore: following convention used above

stvstnfrd, line 266: Looking at this again, I think I need to update the labels earlier on in the helper, maybe somewhere around `self._fix_github_labels()` :wink:

sarina (approved):

PR #147, opened 2021-07-29
feat: fail builds if contributor needs to sign a CLA, but hasn't

stvstnfrd:
and then mark as passing once they have.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,23 +1,47 @@
-FROM python:2.7
-ENV PYTHONUNBUFFERED 1
-ENV PYTHONPATH /app
+FROM ubuntu:focal as app
 
-RUN apt-get update && apt-get install --no-install-recommends -y \
-  bash-completion \
-  exuberant-ctags \
-  && rm -rf /var/lib/apt/lists/*
-RUN echo 'source /usr/share/bash-completion/bash_completion' >> /etc/bash.bashrc
+# System requirements.
+RUN apt-get update && apt-get upgrade -qy
+RUN apt-get install -qy \
+	git-core \
+	language-pack-en \
+	python3.8 \
+	python3-pip \
+	python3.8-dev \
+	libssl-dev
+RUN pip3 install --upgrade pip setuptools
 
-RUN echo 'export HISTFILE=$HOME/.bash_history/history' >> $HOME/.bashrc
+# Python is Python3.
+RUN ln -s /usr/bin/python3 /usr/bin/python
 
-ARG REQUIREMENTS_FILE
-WORKDIR /app
-COPY requirements requirements
-RUN pip install --no-cache-dir -r requirements/${REQUIREMENTS_FILE} && rm -rf /root/.cache
+# Use UTF-8.
+RUN locale-gen en_US.UTF-8
+ENV LANG en_US.UTF-8
+ENV LANGUAGE en_US:en
+ENV LC_ALL en_US.UTF-8
 
-ARG TINI_VERSION
-RUN curl -SL \
-  https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini \
-  -o /tini
-RUN chmod +x /tini
-ENTRYPOINT ["/tini", "--"]
+RUN apt-get install -qy \
+	curl \
+	ca-certificates \
+	gnupg \
+;
+RUN curl https://www.postgresql.org/media/keys/ACCC4CF8.asc \
+    | apt-key add -
+RUN sh -c '\
+    echo "deb http://apt.postgresql.org/pub/repos/apt focal-pgdg main" \
+    > /etc/apt/sources.list.d/pgdg.list \
+'
+RUN apt update -qy
+RUN DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends tzdata
+RUN apt install -qy postgresql-13 libpq-dev
+
+RUN pip install tox
+
+RUN mkdir -p /edx/app/openedx-webhooks
+WORKDIR /edx/app/openedx-webhooks
+COPY Makefile ./
+COPY requirements ./requirements
+RUN make install-dev-requirements
+
+COPY . /edx/app/openedx-webhooks
+CMD ["make", "test"]
--- a/openedx_webhooks/github/dispatcher/actions/test/test_github_cla.py
+++ b/openedx_webhooks/github/dispatcher/actions/test/test_github_cla.py
@@ -1,0 +1,154 @@
+"""
+Confirm expected behavior of the CLA-related functionality
+"""
+# Disable these warnings due to how the fixtures are constructed and used.
+# pylint: disable=redefined-outer-name,unused-argument
+import pytest
+
+from openedx_webhooks.github.dispatcher.actions.utils import update_commit_status_for_cla
+
+
+COMMIT_SHA = 'deadbeefcafe'
+PULL_REQUEST = {
+    'base': {
+        'repo': {
+            'full_name': 'edx/openedx-webhooks-test',
+        },
+    },
+    'number': 1,
+}
+
+
+@pytest.fixture
+def patch_get_commit(mocker):
+    """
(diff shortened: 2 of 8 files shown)


Dockerfile
```diff
@@ -1,23 +1,50 @@
-FROM python:2.7
-ENV PYTHONUNBUFFERED 1
-ENV PYTHONPATH /app
+FROM ubuntu:focal as app
```
stvstnfrd, line 1: I used this `Dockerfile` to run tests locally, as notably, the app requires a PostreSQL installation.

This could be excluded from the PR, though I did find it helpful.
I'm admittedly not sure what the previous `Dockerfile` was trying to do...

This file was cribbed from another edx repo (discovery or registrar; I forget now).

Dockerfile
```diff
@@ -1,23 +1,50 @@
...
+
+RUN mkdir -p /edx/app/openedx-webhooks
+WORKDIR /edx/app/openedx-webhooks
+COPY Makefile ./
+COPY requirements ./requirements
+RUN make install-dev-requirements
+
+COPY . /edx/app/openedx-webhooks
+# CMD ["make", "test"]
+CMD ["py.test", "-rxefs", "--cov=openedx_webhooks", "--cov=tests", "--cov-report=", "--cov-context=test", "openedx_webhooks/github/dispatcher/actions/test/test_github_cla.py", "-k", "TestCla"]
```
stvstnfrd, line 50: TODO: If we keep the dockerfile, delete this line and uncomment the above.

Dockerfile
```diff
@@ -1,23 +1,50 @@
...
+ENV LANGUAGE en_US:en
+ENV LC_ALL en_US.UTF-8
 
-ARG TINI_VERSION
-RUN curl -SL \
-  https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini \
-  -o /tini
-RUN chmod +x /tini
-ENTRYPOINT ["/tini", "--"]
+RUN apt-get install -qy \
```
stvstnfrd, line 25: TODO: Combine this with `apt-get` calls above?

openedx_webhooks/github/dispatcher/actions/test/test_github_cla.py
```diff
@@ -0,0 +1,139 @@
+"""
+Confirm expected behavior of the CLA-related functionality
+"""
+# Disable these warnings due to how the fixtures are constructed and used.
+# pylint: disable=redefined-outer-name,unused-argument
```
stvstnfrd, line 5: Note that pylint doesn't run against this repo, so there are myriad violations.
This file, however, is clean.

openedx_webhooks/github/dispatcher/actions/utils.py
```diff
@@ -15,3 +23,63 @@ def find_issues_for_pull_request(jira, pull_request_url):
...
+    sha = commit.get('sha')
+    logger.debug(f"CLA: SHA %s", sha, repo_name_full)
+    return sha
+
+
+def _get_latest_commit_for_pull_request_data(repo_name_full: str, number: int) -> str:
+    """
+    Lookup the HEAD commit SHA for a pull request
+    """
+    url = f"https://api.github.com/repos/{repo_name_full}/pulls/{number}/commits"
```
stvstnfrd, line 42: It looks like most of these types of functions (hitting the Github API) are actually located right inside `pr_tracking.py`.
Should I just move these there too? This had seemed like a reasonable place to stick them :shrug:

tests/test_rescan.py
```diff
@@ -193,7 +197,6 @@ def test_rescan_blended(reqctx, fake_github, fake_jira):
     assert "changed" not in ret
 
     # We shouldn't have made any writes to GitHub or Jira.
-    fake_github.assert_readonly()
```
stvstnfrd, line 196: Rescans of blended PRs are no longer guaranteed to be read-only, as we may be `POST`ing to the status endpoint, even if the other state is unchanged.

I'm _assuming_ this is a reasonable change of expectations, but open to having missed what this is actually checking, so let me know...

openedx_webhooks/tasks/pr_tracking.py
```diff
@@ -386,6 +387,7 @@ def fix(self) -> None:
         if fix_comment:
             self._fix_bot_comment(comment_kwargs)
         self._add_bot_comments()
+        self.actions.update_commit_status_for_cla(pull_request=self.pr)
```
stvstnfrd, line 390: This bit, the `self.actions...` part, seems to be how we're performing dry-runs?

stvstnfrd (commented):
Pre-review thoughts:

openedx_webhooks/github/dispatcher/actions/utils.py
```diff
@@ -15,3 +23,63 @@ def find_issues_for_pull_request(jira, pull_request_url):
...
+    logger.debug(f"CLA: GET %s", url)
+    response = get_github_session().get(url)
+    log_check_response(response)
+    data = response.json()
+    logger.debug(f"CLA: GOT %s", data)
+    return data
+
+
+def _update_commit_status_for_cla(url, payload):
+    logger.debug(f"CLA: POST %s %s", url, payload)
```
stvstnfrd, line 52: TODO: Remove interpolated string where not needed.

openedx_webhooks/github/dispatcher/actions/utils.py
```diff
@@ -15,3 +23,94 @@ def find_issues_for_pull_request(jira, pull_request_url):
...
+    """
+    logger.info("CLA: GET %s", url)
+    response = get_github_session().get(url)
+    log_check_response(response)
+    data = response.json()
+    logger.info("CLA: GOT %s %s", url, data)
+    cla_status = [
+        status
+        for status in data
+        if status.get('context') == 'cla'
```
nedbat, line 66: Reading between the lines of the GitHub API, the context needs to be unique across all possible checkers.  I see contexts like "jenkins/quality".  We should change "cla" to be more specific to us, like "edx/cla" at least.

openedx_webhooks/github/dispatcher/actions/utils.py
nedbat: It's fine to have them here.  the repo is kind of ad-hoc at this point.

openedx_webhooks/tasks/pr_tracking.py
```diff
@@ -774,3 +776,8 @@ def update_labels_on_pull_request(self, *, labels: List[str]) -> None:
         logger.info(f"Patching labels on PR {self.prid}: {labels}")
         resp = get_github_session().patch(url, json={"labels": labels})
         log_check_response(resp)
+
+    def update_commit_status_for_cla(self, *, pull_request: PrDict) -> None:
+        result = update_commit_status_for_cla(pull_request)
+        if result is not None:
+            self.happened = True
```
nedbat, line 783: This doesn't match up with what `update_commit_status_for_cla` returns: it returns True or False for whether the cla is signed or not, even if it didn't have to change the status.

nedbat:
We should add some lines to test_pull_request_opened.py that asserts about the status updates happening.

openedx_webhooks/github/dispatcher/actions/utils.py
stvstnfrd: Done in pending commit.

openedx_webhooks/tasks/pr_tracking.py
stvstnfrd: Good catch. That bit evolved over time and you're correct; they no longer match. I have updated `update_commit_status_for_cla` to return the shape `tuple(has_changed: bool, has_signed: bool)`.

stvstnfrd (commented):
@nedbat I made the changes to return type and context. Currently looking at the test suite to see the best way to check that behavior (open to suggestions).

nedbat (approved):

PR #149, opened 2021-08-18
fix: protect all pages with basic auth

nedbat:
Also, use the correct GET view names in the HTML template, although the
POST view names worked also, since it was only getting the URL, which is
the same for both.


stvstnfrd (approved):
Code/Auth changes look good!

(not familiar w/ the scriv/changelog stuff yet, but willing to defer for the sake of this PR)

PR #150, opened 2021-08-18
fix: cleanup JIRA session management

stvstnfrd:
- to avoid a potential import-time network call
- to remove complexity of JIRA client injection


openedx_webhooks/lib/jira/client.py
```diff
@@ -15,13 +15,16 @@
...
-_oauth_info = dict(
-    access_token=environ_get('JIRA_ACCESS_TOKEN'),
-    access_token_secret=environ_get('JIRA_ACCESS_TOKEN_SECRET'),
-    consumer_key=environ_get('JIRA_OAUTH_CONSUMER_KEY'),
-    key_cert=b64decode(environ_get('JIRA_OAUTH_PRIVATE_KEY')),
-)
-
-# (jira.JIRA): An authenticated JIRA API client session
-jira_client = JIRA(_server, oauth=_oauth_info)
+def get_authenticated_session() -> JIRA:
```
nedbat, line 18: When I was puzzling through the Jira authentication, it took me a bit to realize that this was a `JIRA` object, and the other was a `requests.Session` object with authentication for Jira.  I'm wondering if there's a better name for this that doesn't use the word "session".

nedbat:
I broke the tests... the fix is in #151.

stvstnfrd:
@nedbat I have incorporated your changes from #151 here (via `git cherry-pick`);
we can close that in favor of this.

nedbat (approved):

PR #155, opened 2021-09-15
try out commitlint.

nedbat:


commitlint.config.js
```diff
@@ -0,0 +1,25 @@
...
+      [2, 'always', [
+        'revert', 'feat', 'fix', 'perf', 'docs', 'test', 'build', 'refactor', 'style', 'chore', 'temp'
+      ]],
+  },
+
+  ignores: [
+    // Allow GitHub revert messages, like:
+    //    Revert "introduce a bug"
+    //    Revert "introduce a bug" (#1234)
+    message => /^Revert ".*"( \(#\d+\))?/.test(message)
```
feanil, line 23: Sometimes people merge PRs into their PRs in which case there would be merge commits in a PR, do we want to also ignore that or ask people to update those commit messages?

nedbat: commitlint has default ignores, though the specifics are not documented.  They include merge messages: https://github.com/conventional-changelog/commitlint/blob/master/%40commitlint/is-ignored/src/defaults.ts

nedbat:
This has moved to https://github.com/edx/edx-lint/pull/199

PR #158, opened 2021-12-15
Feat: Add salesforce provided CSV priority support

Carlos-Muniz:
Previously, the bot used a dictionary of users made from `people.yaml`.
In an effort to be more secure, the CLA bot will no longer be depending
on _just_ the `people.yaml` file.

Now, the CLA bot depends on `salesforce-export.csv`. Due to the csv
file lacking important information relating to organization, the
`people.yaml` file will be used to supplement the csv. The csv will take
priority, and users mentioned in the yaml file but not the csv file, will not
be considered by the CLA bot.

This functionality has been tested with the test `test_check_csv_priority()` which uses a new entry in tests/repo_data/people.yaml.


openedx_webhooks/info.py
```diff
@@ -7,6 +7,7 @@
 from typing import Dict, Iterable, Optional, Union
 
 import yaml
+import csv
```
nedbat, line 10: nit: stdlib imports should be in the section above, and in alphabetical order (before "import datetime")

openedx_webhooks/info.py
```diff
@@ -25,6 +26,12 @@ def _read_repotools_yaml_file(filename):
     """Read a YAML file from the repo-tools-data repo."""
     return yaml.safe_load(_read_repotools_file(filename))
 
+def _read_repotools_csv_file(filename):
+    """
+    Read a CSV file from the repo-tools-data repo. Returns a list of rows.
```
nedbat, line 31: You could mention here that the first row is assumed to be headers which are skipped.
I know you don't want to fiddle with csv any more, but DictReader might be more convenient.

openedx_webhooks/info.py
```diff
@@ -35,7 +42,38 @@ def _read_repotools_file(filename):
     return resp.text
 
 def get_people_file():
-    return _read_repotools_yaml_file("people.yaml")
+    """
+    Returns people in the formatted dictionary.
```
nedbat, line 46: "the formatted dictionary" sounds odd?

openedx_webhooks/info.py
```diff
@@ -35,7 +42,38 @@ def _read_repotools_file(filename):
...
+        github_username: {
+            name: "",
+            agreement: "",
+            institution: ""
+            jira: "",
+            other_emails: [...],
+            committer: {...},
+            comments: [...],
+            before: {...}
+        }
```
nedbat, line 57:
```suggestion
        },
        ...
```

tests/test_info.py
```diff
@@ -195,3 +195,9 @@ def test_is_draft_pull_request(fake_github, title):
     # No matter what the title, a pr is Draft if it says it is.
     pr = fake_github.make_pull_request(title=title, draft=True)
     assert is_draft_pull_request(pr.as_json())
+
+def test_check_csv_priority():
+    people = get_people_file()
+    user = 'Carlos-Muniz'
+    # This user exists in the yaml but not in the csv, and should not be in people
+    assert people.get(user) == None
```
nedbat, line 203: The last line is missing a newline.  Is your editor configured to read .editorconfig files? It should be, to automatically enforce some rules like this.

nedbat (changes requested):

nedbat:
You can add a changelog entry with `scriv create`. It will make a new file that you can commit with this pull request, and it will be collected into the changelog when the release is done.

openedx_webhooks/info.py
Carlos-Muniz: Moved!

openedx_webhooks/info.py
Carlos-Muniz: Added!

openedx_webhooks/info.py
```diff
@@ -35,7 +42,38 @@ def _read_repotools_file(filename):
...
+        if acct_name == "Individual Contributors":
+            people[github_username]["agreement"] = 'individual'
+        else:
+            people[github_username]["agreement"] = 'institution'
+            people[github_username]["institution"] = acct_name
+    
+    people_data_yaml = _read_repotools_yaml_file("people.yaml")
+    for p in people:
+        people[p].update(people_data_yaml[p])
```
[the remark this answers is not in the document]
Carlos-Muniz: Yes. You're right. I'll add a test and implement code that makes sure that this test passes.

tests/test_info.py
Carlos-Muniz: Test and implementation has been added. 
And I've added the .editorconfig functionality to my editor

openedx_webhooks/info.py
Carlos-Muniz: I've reworded this comment

openedx_webhooks/info.py
```diff
@@ -35,7 +42,50 @@ def _read_repotools_file(filename):
...
+            name: "",
+            agreement: "",
+            institution: ""
+            jira: "",
+            other_emails: [...],
+            committer: {...},
+            comments: [...],
+            before: {...},
+            ...
+        }
```
nedbat, line 58: This got put in the wrong place.  The comma-dot-dot-dot should be after the dict of github_username, to indicate that there are many usernames in the result.

openedx_webhooks/info.py
```diff
@@ -25,6 +25,13 @@ def _read_repotools_yaml_file(filename):
     """Read a YAML file from the repo-tools-data repo."""
     return yaml.safe_load(_read_repotools_file(filename))
 
+def _read_repotools_csv_file(filename):
+    """
+    Read a CSV file from the repo-tools-data repo. Returns a list of rows. The
```
nedbat, line 30: Does it return a list of rows?

tests/test_info.py
```diff
@@ -195,3 +195,14 @@ def test_is_draft_pull_request(fake_github, title):
     # No matter what the title, a pr is Draft if it says it is.
     pr = fake_github.make_pull_request(title=title, draft=True)
     assert is_draft_pull_request(pr.as_json())
+
+def test_check_csv_users_only():
+    people = get_people_file()
+    user = 'Carlos-Muniz'
+    # This user exists in the yaml but not in the csv, and should not be in people
+    assert people.get(user) == None
```
nedbat, line 203: For whatever reason, it's idiomatic in Python to use `is None` rather than `== None`.

openedx_webhooks/info.py
```diff
@@ -35,7 +42,50 @@ def _read_repotools_file(filename):
...
+            people[github_username]["agreement"] = "none"
+        else:
+            people[github_username]["agreement"] = 'institution'
+            people[github_username]["institution"] = acct_name
+
+    people_data_yaml = _read_repotools_yaml_file("people.yaml")
+    for p in people:
+        # Prioritzes the csv by first updating the yaml's values with csv values
+        # And then updates any missing dict fields using yaml's fields
+        people_data_yaml.get(p).update(people[p])
```
nedbat, line 86: This .get will produce None if the key is missing, which will then fail the `.update`.  You can't assume the keys in the csv all exist in the yaml, so this might throw an exception?  There's an idea for a test.

nedbat: BTW, if you do know that the key will be in the dict, then you don't need .get(), and can just use indexing, like in the next line.

tests/test_info.py
```diff
@@ -195,3 +195,14 @@ def test_is_draft_pull_request(fake_github, title):
...
+
+def test_check_csv_org_priority():
+    people = get_people_file()
+    user = 'mmprandom'
+    assert people.get(user).get('agreement') == 'individual'
```
nedbat, line 208: Here's another place where `[]` is better than `.get()` (twice).

tests/test_info.py
nedbat: I'm wondering if there's a way to assert that our test data is still what we think it is.  For example, your comment says you are in the yaml but not the csv.  Someone might unthinkingly edit the test data and break that assumption.  But this is probably beyond the scope of this pull request.

nedbat (changes requested):

openedx_webhooks/info.py
Carlos-Muniz: Yup, I added in the wrong place. Fixed!

openedx_webhooks/info.py
Carlos-Muniz: Clarified this comment.

tests/test_info.py
Carlos-Muniz: Changed from == to is

openedx_webhooks/info.py
Carlos-Muniz: In the case that a user is in the csv, but not in the yaml, we don't want to throw an exception. We would prefer to just have the details we have from the csv. It seems like putting the dictionary updates within an if-statement checking that a user is in the yaml would bypass that.

Carlos-Muniz: @nedbat But how would we test this? Is it enough to have a user in the test csv but not in the test yaml, and make sure that there aren't any yaml specific keys in people dictionary for that user?

tests/test_info.py
Carlos-Muniz: Changed.

openedx_webhooks/info.py
nedbat: Yes, it's enough to add a person into the csv or yaml in the way that you want.  That's what the test data is for: it's only for our tests, so we can make it as weird and arbitrary as we need to.

Carlos-Muniz: Fix and test have been added

nedbat (approved):

nedbat:
Don't forget to squash the commits down to one.

PR #162, opened 2022-01-21
When PR is reopened, reopen the Jira ticket

sarina:
on action `reopened` we want to reopen the PR

my thought it should be reopened in Needs Triage, this accounts for users reopening on their own and we will need to look at them again

### DIFF
--- a/tests/fake_github.py
+++ b/tests/fake_github.py
@@ -162,6 +162,14 @@ class PullRequest:
         self.merged = merge
         self.closed_at = datetime.datetime.now()
 
+    def reopen(self):
+        """
+        Re-open a pull request.
+        """
+        self.state = "open"
+        self.merged = False
+        self.closed_at = None
+
     def add_comment(self, user="someone", **kwargs) -> Comment:
         comment = self.repo.make_comment(user, **kwargs)
         self.comments.append(comment.id)
--- a/openedx_webhooks/bot_comments.py
+++ b/openedx_webhooks/bot_comments.py
@@ -191,7 +191,6 @@ def github_end_survey_comment(pull_request: PrDict) -> str:
     """
     Create a "please fill out this survey" comment.
     """
-    print(pull_request)
     is_merged = pull_request.get("merged", False)
     url = SURVEY_URL.format(
         repo_full_name=pull_request["base"]["repo"]["full_name"],
--- a/openedx_webhooks/github_views.py
+++ b/openedx_webhooks/github_views.py
@@ -85,6 +85,7 @@ def hook_receiver():
     pr = event["pull_request"]
     pr_number = pr["number"]
     action = event["action"]
+    pr["hook_action"] = event["action"]
 
     pr_activity = f"{repo} #{pr_number} {action!r}"
     if action in ["opened", "edited", "closed", "synchronize", "ready_for_review", "converted_to_draft"]:
--- a/openedx_webhooks/tasks/pr_tracking.py
+++ b/openedx_webhooks/tasks/pr_tracking.py
@@ -234,7 +234,9 @@ def desired_support_state(pr: PrDict) -> Optional[PrDesiredInfo]:
     else:
         desired.is_ospr = True
 
-    if pr["state"] == "open":
+    if pr.get("hook_action") == "reopened":
+        state = "reopened"
+    elif pr["state"] == "open":
         state = "open"
     elif pr["merged"]:
         state = "merged"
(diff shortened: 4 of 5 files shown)


sarina:
@nedbat before I write any tests, do you think this could be the right approach? I'm basically munging the state when the action is reopened, because the action isn't passed through (PR doesn't have that info). but there's no tests for `hook_receiver` so I can't test that part of it.

nedbat:
I wonder if we need to know that the action was "reopened"?  It might be enough to act on a reopened event, then the rest of the logic see an open PR, and will make everything right?

sarina:
@nedbat happy to try that. I'm not sure I can write good tests for that though. Are you saying to basically have the lines

```
    if action == "reopened":
        pr["state"] = "reopened"
```

and leave it at that?

nedbat:
Yes, I meant leave it at that.  I'm trying to think through what else would really be needed?  The code is written to make everything right regardless of the current state, so it might be enough.

sarina:
@nedbat I agree on that. I pushed the changes, although I'm not sure there's a way to write a test here.

tests/fake_github.py
```diff
@@ -162,6 +162,14 @@ def close(self, merge=False):
         self.merged = merge
         self.closed_at = datetime.datetime.now()
 
+    def reopen(self):
+        """
+        Re-open a pull request.
+        """
+        self.state = "open"
```
sarina, line 169: @nedbat I got the transitions working by setting this to "reopened", and then being able to use that info within `desired_support_state` to set the right jira status (we need that info in that function to set the new jira state properly). it will fail the `assert anything_happened2 is False` test because the Jira status has changed.

I could either do this pretty simple thing (munge the state if the action is reopened), or, make this test pass with state="open" (I can think of a way to do that, which is within `desired_support_state` for every "open" pr, check to see if there is a SURVEY comment). Either way the `assert anything_happened2 is False` will not be true, because the jira status is changing.

Thoughts on best way?

sarina: Also I think the state of the PR has always been "open" when the action is "reopened". I can't find an exact reference, but do a command line reopen you just make an api call with state=open.

nedbat: It's fine to change the test "assert anything_happened" things to be correct for the way the code behaves.  The thing to do first is figure out what we want to happen, and then make it happen.  This line in fake_github has to stay as it is, because it emulates GitHub: when a PR is reopened, the state is set to "open."

sarina: OK, so this comes around to my initial problem of not knowing how to test the actual change I want to make. Sure, this comes in with `state = "open"`. It _also_ comes in with `event["action"] = "reopened"`.

I'm suggesting to intercept that action in `github_views.hook_receiver()` and munge the state to `"reopened"`. 

```
    pr = event["pull_request"]
    pr_number = pr["number"]
    action = event["action"]
    if action == "reopened":
        pr["state"] = "reopened"
```

That way, code in `pr_tracking.desired_support_state` understands this is a different case and can, around line 305, add a clause

```
        if state == "closed":
            desired.jira_status = "Rejected"
        elif state == "merged":
            desired.jira_status = "Merged"
        elif state == "reopened":
            desired.jira_status = "Community Manager Review"
```

But, I don't see a way in the tests to pass the action context. In `fake_github` we mock the PR details, but we don't mock the full event call as far as I can tell. These tests rely on setting _state_ differences, rather than distinguishing between two different actions that caused the PR to get into the same state.

Check out my latest commit. The tests will fail. But if you manually munge the PR state, they pass. So how could I go about leaving the test definition alone, but making sure the new code at github_views L88-89 get executed properly?

sarina: Hmm I never actually pushed that commit.... one day, we'll get this 😁

sarina:
Hey @nedbat - this has been languishing a bit. If you don't have time to continue on it, I'll go ahead and close.

sarina:
I'm closing this for now. If there's time/space for a review, I think this would be a valuable addition.

nedbat:
I just made you a tweak PR for it the other day: #168.

sarina:
@nedbat this seems good to me. 

to fix the comment issue, I suppose the bot would need to, on a reopen event, find the CLOSED comment and delete it. Is that right? I think it may be better to just get this merged, and work on that separately (assuming the bot has permissions to delete comments)

nedbat:
See #169.

nedbat:
This work has been merged as part of #169.

PR #166, opened 2022-02-09
test: add a failing test of reopen behavior

nedbat:
@sarina Here's an approach to testing re-opening.

This points out two things:

First, the test fails because after closing and re-opening, the Jira ticket is still "Rejected."  I haven't looked into why that is.

Second, closing the PR adds a bot comment to take the survey.  Do we want to do anything about that? We could remove that comment, or we could add another "yay, re-opened!" comment, or something else.


sarina:
Thanks @nedbat - this is truly helpful 💯 

About the comment, removing the comment would be best, but since reopening is by far not a common use case I'm not overly concerned about it.

PR #168, opened 2022-03-16
temp: map the action to a new key in the pr json, for reopening

nedbat:
This makes the tests pass.  Since the "reopen" fact isn't in the pull request, but instead in the hook action, I added it to the pr data as a separate piece of information.

It's not great that the hook handler isn't tested by the test suite, and we have to add a line to the tests to parallel a line in the handler, but I think this will work.

@natabene is this the behavior you want?: When a pull request is reopened, the Jira status becomes "Community Manager Review", regardless of other considerations.

### DIFF
--- a/openedx_webhooks/github_views.py
+++ b/openedx_webhooks/github_views.py
@@ -85,8 +85,7 @@ def hook_receiver():
     pr = event["pull_request"]
     pr_number = pr["number"]
     action = event["action"]
-    if action == "reopened":
-        pr["state"] = "reopened"
+    pr["hook_action"] = event["action"]
 
     pr_activity = f"{repo} #{pr_number} {action!r}"
     if action in ["opened", "edited", "closed", "synchronize", "ready_for_review", "converted_to_draft", "reopened"]:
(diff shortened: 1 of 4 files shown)


nedbat:
BTW, another not-great behavior here: but probably not a big deal: when the PR is closed, the bot comments, "even though your PR wasn't merged, fill out the survey."  Then when the PR is eventually merged, it doesn't add a congratulations comment.

nedbat:
You can see a PR that went through this cycle:  https://github.com/nedbat/webhook-testing/pull/99

natabene:
> This makes the tests pass. Since the "reopen" fact isn't in the pull request, but instead in the hook action, I added it to the pr data as a separate piece of information.
> 
> It's not great that the hook handler isn't tested by the test suite, and we have to add a line to the tests to parallel a line in the handler, but I think this will work.
> 
> @natabene is this the behavior you want?: When a pull request is reopened, the Jira status becomes "Community Manager Review", regardless of other considerations.

Yes

sarina:
This makes sense. The PR comment is lame but not worth fixing I think right now.

PR #170, opened 2022-04-03
feat: Point the openedx_webhooks bot to the new data repo.

feanil:
As a part of the transition from edX to tCRIL, the repo-tools-data repo did
not get transferred.  A copy of the data needed by this application was made
and put into the openedx-webhooks-data repo in the openedx org.

This change points the app to load data from that repo instead.  Operationally,
that repo has the credentials for the edx-webhooks GitHub user and so that user
has also been given access to the new openedx-webhooks-data repo.

### DIFF
--- a/openedx_webhooks/info.py
+++ b/openedx_webhooks/info.py
@@ -38,7 +38,7 @@ def _read_repotools_file(filename):
     Read the text of a repo-tools-data file.
     """
     github = get_github_session()
-    resp = github.get(f"https://raw.githubusercontent.com/edx/repo-tools-data/master/{filename}")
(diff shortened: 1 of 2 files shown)


feanil:
@Carlos-Muniz I can do that as a follow-up,  I want this to be the minimal change to validate that the data is being pulled from the right place so I'd like to make those changes as a follow-up PR, I've made https://github.com/openedx/openedx-webhooks/issues/171 to do that follow-up so it doesn't get lost.

Carlos-Muniz (approved):
Thanks for addressing my comment!

PR #174, opened 2022-04-05
refactor: Rename the edx_repo_tools_data lib to webhooks_data

feanil:


Carlos-Muniz (approved):

nedbat:
Why `feat!` commit rather than `refactor`?  Aren't all of the callers of that moved code in this repo?

feanil:
yea, that's a good point, I was going back and forth between the two and ended on feat! because I wasn't sure what else might be using it.

PR #177, opened 2022-04-06
fix: re-label to any jira status

nedbat:
We weren't paginating the labels, so we only got 30.  In edx-platform,
"open edx community review" was the 31st label, so we thought it didn't
exist, so we wouldn't add the label when a Jira ticket was transitioned
to Open edX Community Review.

### DIFF
--- a/openedx_webhooks/tasks/github_work.py
+++ b/openedx_webhooks/tasks/github_work.py
@@ -2,6 +2,8 @@
 Operations on GitHub data.
 """
 
+from typing import Any, Dict
+
 from openedx_webhooks.info import get_labels_file
 from openedx_webhooks.oauth import get_github_session
 from openedx_webhooks.tasks import logger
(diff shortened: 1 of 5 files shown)


openedx_webhooks/tasks/github_work.py
```diff
@@ -12,12 +14,19 @@
...
+    return repo_labels
+
+
 @memoize_timed(minutes=15)
 def synchronize_labels(repo: str) -> None:
     """Ensure the labels in `repo` match the specs in {DATA_FILES_URL_BASE}/labels.yaml"""
 
     url = f"/repos/{repo}/labels"
-    repo_labels = {lbl["name"]: lbl for lbl in paginated_get(url, session=get_github_session())}
+    repo_labels = get_repo_labels(repo)
```
feanil, line 29: Nice, so we thought about this in one part of the code.

feanil (approved):

openedx_webhooks/tasks/github_work.py
nedbat: Finally, an up-side to "two ways for everything": one of the ways was right :)

PR #180, opened 2022-05-25
style: fix mypy and pylint errors

nedbat:
I guess we don't use setup.py, since it had an extraneous parenthesis?!?


feanil (approved):
Should we remove the setup.py if it's not being used?  What's the value in having it?

nedbat:
We should remove it if it isn't used. I don't know what it used to be used for.  A challenge with this repo is limiting the chasing down of dead stuff like this.

PR #198, opened 2022-09-07
Fix github url strings (org edx -> openedx)

sarina:
## This PR was autogenerated

This pr replaces the old GitHub organization, github.com/edx, with the new GitHub organization, github.com/openedx.

Tagging @openedx/tcril-engineering for review, but others are welcome to provide review.

Ref: https://github.com/openedx/tcril-engineering/issues/42


.editorconfig
```diff
@@ -2,7 +2,7 @@
 # ** DO NOT EDIT THIS FILE **
```
nedbat, line 2: This comment is real: you can't just change this file (or pylintrc) without affecting it in the future.  These files end with a checksum that is now invalid.

sarina:
@nedbat I ran `edx_lint` on pylintrc & .editorconfig but the diff is way huger than I expected - here was the diff for django-wiki: https://github.com/openedx/django-wiki/pull/144/commits/ad061b6009340b53330e6633aeffec0dab0a82b5

Any thoughts?

sarina:
@nedbat figured out the tests, and I've seen similarly long diffs now in `pylintrc` to realize it's just yours here was kinda far behind. I think everything here is set.

tests/test_pull_request_opened.py
```diff
@@ -276,14 +276,15 @@ def test_core_committer_pr_opened(has_jira, sync_labels_fn, fake_github, fake_ji
...
-    sync_labels_fn.assert_called_once_with(f"edx/edx-platform")
+    sync_labels_fn.assert_called_once_with(f"openedx/edx-platform")
 
     # Check the GitHub comment that was created.
     pr_comments = pr.list_comments()
     assert len(pr_comments) == 1
     body = pr_comments[0].body
     check_issue_link_in_markdown(body, issue_id)
     assert "Thanks for the pull request, @felipemontoya!" in body
+    print(f"***************BODY IS:\n{body}")
```
nedbat, line 287: This should be removed.

tests/payloads/github.json
```diff
@@ -4,9 +4,9 @@
   "pull_request": {
     "url": "https://api.github.com/repos/edx/openedx-webhooks/pulls/82",
```
nedbat, line 5: It kind of doesn't matter, since these .json files are just examples saved for us to use while developing, but here's an `/edx/` in the api.github.com URL.  There are other `/edx/` throughout the JSON.

But I wonder if there are other api.github.com URLs that need to be fixed in other repos where they are really used for something?

tests/test_pull_request_opened.py
sarina: 😳

tests/payloads/github.json
sarina: HMM yes. Probably. Time for another set of prs 😆 😱 

Makes sense to change them here, at least.

nedbat: Looking through the repos, it looks like the only GitHub api URLs that need fixing are in repo-health.

tests/test_pull_request_opened.py
nedbat, line 287:
```suggestion
```

sarina:
Pushed a new commit about the api string; oddly the command I'd been successfully using for days now does not want to work. I don't love computers.

nedbat (approved):

PR #204, opened 2022-10-04
docs: Fix the spacing.

feanil:
Because of the trimming in the filters, we have to explicitly add a
newline to separet these two paragraphs.


feanil:
Tested here: https://github.com/nedbat/webhook-testing/pull/122

sarina (approved):

PR #206, opened 2022-10-07
feanil/cleanup unused code

feanil:
- docs: Fix the spacing.
- feat!: Remove the other_emails field.
- feat!: Remove the `jira` field.
- feat!: Remove core committer champion pings on PRs.

### DIFF
--- a/openedx_webhooks/bot_comments.py
+++ b/openedx_webhooks/bot_comments.py
@@ -32,7 +32,6 @@ class BotComment(Enum):
     CORE_COMMITTER = auto()
     BLENDED = auto()
     OK_TO_TEST = auto()
-    CHAMPION_MERGE_PING = auto()
     END_OF_WIP = auto()
     SURVEY = auto()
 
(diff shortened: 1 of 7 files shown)


nedbat:
I don't see `docs: fix the spacing` in the changes.

nedbat:
Should we update repo-tools-data-schema to get rid of fields we aren't using anymore?

nedbat:
I wonder about `other_emails`: that was because we'd know their non-work email first, and then get edx.org emails, and we didn't want to lose information. Also, it helped figure people out if they were committing from different GitHub accounts.

But we aren't maintaining it anymore, so I guess it should go.

nedbat (approved):
The changes seem good as far as they go, but what's the plan for repo-tools-data-schema?

feanil:
We'll have to update repo-tools-data-schema in the same way I guess.  I added https://github.com/openedx/repo-tools-data-schema/issues/18 to follow up in that repo.

PR #208, opened 2022-10-27
Feanil/update to python 3.10

feanil:
This is a pre-req for updating to the latest heroku stack.

Related: https://github.com/openedx/openedx-webhooks/issues/205


feanil:
The dependencies are out of date and a bit of a mess which are making this update more complicated.  I think I'll need to clean some of that up first before we attempt this.

PR #211, opened 2022-11-02
feanil/fix project queries

feanil:
- revert: "temp: Disable trying to add new issues to github projects."
- feat: Update github GraphQL queries to use the v2 graphql.
- test: Update fake_github.py to return projectV2 data.


nedbat:
For completeness, there's also mention of projectNext in tests/test_helpers.py that could be changed to avoid false positives when searching the code.

nedbat (approved):

feanil:
Tested via staging here: https://github.com/openedx/mdrst/pull/29

We also had to give the webhooks token the `project` scope which it did not have before.

PR #214, opened 2022-11-03
fix: Check to see if the Jira Server setting is set.

feanil:
Flask Dance's authorized function seems to be returning true even if we
don't provide any Jira settings. Then we fail to actually get info about
ourselves:

requests.exceptions.MissingSchema: Invalid URL '/rest/api/2/myself': No scheme supplied. Perhaps you meant http:///rest/api/2/myself?

Don't try to do Jira things if the JIRA_SERVER setting is not set.


openedx_webhooks/ui.py
```diff
@@ -28,7 +28,7 @@ def index():
 
 
     jira_username = None
-    if jira_session.authorized:
+    if jira_session.authorized and settings.JIRA_SERVER:
```
nedbat, line 31: This check looks backwards, but I guess `.authorized` here might mean, `connected anonymously to jira`?

feanil: Yea, I'm not sure, I can reverse them but right now `jira_session.authorized` is returning `True` even when there are no jira settings set  and I'm not really sure how deep to dig into flask-dance to figure out what's going on.

nedbat (approved):

PR #216, opened 2022-11-08
fix: Get more commits by default.

feanil:
The commits api only gets 30 commits.  Up the default number of commits
we get to 100 in case someone has a long running branch.

Long term we should add paging to this endpoint but this should fix it
in most cases.  The function currently returns the full JSON payload so
we'll need to update the return signature if we want to start using
paging here.  This may impact callers so we defer that work for now.


Carlos-Muniz (approved):
LGTM :rocket:

feanil:
https://github.com/openedx/openedx-webhooks/issues/217 should deal with this in the long term.

sarina:
getting ready to make my 101st commit on my pr... 😂

PR #225, opened 2023-01-27
Clean ups

nedbat:
A number of clean-ups, particularly reducing pylint and mypy noise.


feanil (approved):
uffh, thanks for doing all of this!

PR #233, opened 2023-04-14
temp: more logging to see jira data being sent

nedbat:


feanil (approved):
Seems fine assuming the payload will not have any secrets/credentials in it.

nedbat:
The logging is temporary and will help us see if there are secrets.

PR #245, opened 2023-06-17
feat!: stop syncing labels from `labels.yaml`

kdmccormick:
The repo_checks[1] tool now does this, and we don't want two systems simultaneously trying to create or delete labels.

Note that the bot will still *add* and *delete* a handful of labels from *PRs* (`open-source-pull-request`, `core-committer`, etc). What this commit changes is that the bot will not longer *create* nor *delete* labels from the *entire repository*.

The `labels.yaml` file is now unused and can be safely deleted from any openedx-webhooks data repositories.

[1] https://github.com/openedx/repo-tools/tree/master/edx_repo_tools/repo_checks

Part of: https://github.com/openedx/openedx-webhooks/issues/218


nedbat:
I think it might be true that some of fake_github.py can also be removed now, since it emulates label REST API endpoints that we no longer use.  It's not critical to remove it, but it will be unused code.

nedbat:
This has been covered in #249

PR #246, opened 2023-06-17
feat!: stop adding `core committer` label

kdmccormick:
Stopped the bot from adding the `core committer` GitHub label
to pull requests to repos on which the bot believes the author
to have write access. The bot's data source for repository access,
`people.yaml`, is outdated, we do not yet have a strategy for
keeping it updated. Until further notice, coding Core Contributors
are asked to add the `core contributor` label to their pull requests
manually.

Another implication of `people.yaml` being oudated is that
the bot is not correctly choosing when to create Jira tickets,
which welcome/help message to post, etc. We are not addressing
that issue currently.

Part of: https://github.com/openedx/openedx-webhooks/issues/227

PR #250, opened 2023-07-26
Finish #246: feat!: stop adding core committer label

nedbat:
I only had to remove two lines from a test.


kdmccormick (approved):
Nice! Would you be able to announce the change?

nedbat:
Sure.

PR #251, opened 2023-08-08
Remove history, core contributor status, and people.yaml

nedbat:
This removes a lot...


feanil (approved):
I was a little hesitant to remove the core_committer PR comment since I think there is a version of that which could be useful but given that we'd have to do some work in SF to start tracking that info again, I don't think it's useful to keep for now.  We don't have a plan to track that data there any time soon since it adds a bunch of on-going overhead.

nedbat:
One option we could explore: keep people.yaml specifically for the core contributor status.  It works now, and doesn't need to be updated that often.

nedbat:
OK, as discussed in Slack, we aren't going to keep people.yaml, which has a more complicated schema than we need anyway.   I don't know when we might get CC data from Salesforce, but when we do, we can put back some parts of this code.  It feels a little odd to remove it knowing that we might want it back in the future, but we know where it is and we can come back to get it.

PR #253, opened 2023-08-24
feat!: remove the survey link

kdmccormick:
Before merging, need to confirm both of these suspicions: 

* Admin access to the Google form (and thus its results) were lost during the Axim/2U decoupling and subsequent edX->2U Google migration.

* There is currently no desire by either Axim or 2U to recreate and manage the survey.


nedbat:
I'm asking if the form and data can be rescued.

nedbat:
(a test comment)

nedbat:
(this is another test comment.)

nedbat:
2U IT reports that they cannot retrieve the old form or data.

kdmccormick:
Thanks Ned. I'll ask the team today whether we want to make a new form, or just nix it.

kdmccormick:
Good news--we found the data, and it's still coming in, so turning off the survey is less urgent than I had thought.

I wrote up a follow-up task for the future: https://github.com/openedx/openedx-webhooks/issues/259

PR #254, opened 2023-08-28
feat: commenting on a PR triggers updates and re-checks.

nedbat:
Ed asked for a simpler way to re-check the CLA than editing the title of the PR.  Now any comment on a PR will re-check the PR.  No special syntax needed.


feanil (approved):
This seems fine, I'm a little concerned about the significant increase in the number of events processed so it'd be good to keep an eye on the bot after the PR is deployed and make sure it doesn't fall over from the increased number of items it has to process.  If it's fine through an afternoon it will probably mid week, it will probably be fine.

nedbat:
The bot has be receiving these events for a while, but ignoring them.  The load will be on the celery queues.  I'll deploy it tomorrow and keep an eye on it.

nedbat:
I've deployed it (this comment is a test...)

PR #258, opened 2023-08-31
Massive deletion of code I didn't like and we didn't use.

nedbat:


feanil (approved):
Jira removal makes sense, I was surprised by the amount of the github dispatch code that's duplicated.

nedbat:
@feanil There don't seem to be any settings that were used only by this deleted code.

PR #261, opened 2023-09-13
build: remove codecov

nedbat:
It only ever complained, and we weren't looking at the results.


nedbat:
OK, fine. I feel like I should know how to do this without a third-party service, but it's not at my fingertips, so I'll do something more valuable right now :)

PR #263, opened 2023-09-15
refactor!: no longer create Jira issues when pull requests are created

nedbat:
I tried to remove code that we wouldn't need in the future.  Some basic Jira operation code is still here, but not called.  There is probably more code that can be deleted, but this was a lot already.

### DIFF
--- a/openedx_webhooks/tasks/pr_tracking.py
+++ b/openedx_webhooks/tasks/pr_tracking.py
@@ -45,8 +45,6 @@ from openedx_webhooks.info import (
     is_draft_pull_request,
     is_internal_pull_request,
     is_private_repo_no_cla_pull_request,
-    jira_project_for_blended,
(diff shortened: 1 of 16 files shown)


openedx_webhooks/tasks/pr_tracking.py
```diff
@@ -473,9 +407,6 @@ def _make_jira_issue(self) -> None:
         Make our desired Jira issue.
```
feanil, line 407: Shouldn't this whole function go away?  What's the reason for that,  I don't quite see how you've sliced the code.

feanil (commented):
I assume the rest can be removed once the update_jira_issue code is removed?

feanil (approved):
Based on our conversation "in-person" I think this is a fine line and we're good with the lowered coverage for now.

PR #265, opened 2023-09-27
Labels trigger Jira issue creation

nedbat:
This has the basic functionality, except:
- The project is hardcoded to "TODO"
- I think we need an environment variable for where to find jira-info.yaml, so that staging and production can differ.
- There's no error handling if someone labels a PR "jira:bogus"

The commits are roughly partitioned into separate work, but I had to re-think some things about how the jira ticket is stashed in the comments, and I didn't go back to try to make it a straight path.


openedx_webhooks/settings.py
feanil, line 1: I'm curious about why you didn't just stick to the module namespace.  That's a pattern common in Django so it was familiar.  What you've done is not complicated but I'm just trying to understand what lead you in this direction?  Especially once you pull the test settings into their on settings file.

openedx_webhooks/tasks/pr_tracking.py
```diff
@@ -227,6 +226,10 @@ def desired_support_state(pr: PrDict) -> PrDesiredInfo:
     else:
         state = "closed"
 
+    # A label of jira:xyz means we want a Jira issue in the xyz Jira.
+    label_names = set(lbl["name"] for lbl in pr["labels"])
+    desired.jira_nicks = {name.partition(":")[-1] for name in label_names if name.startswith("jira:")}
+
     desired.jira_initial_status = "Needs Triage"
```
feanil, line 233: This is an assumption about what states are expected in any given Jira project that is added. We should probably document this somewhere.  ie.  "Setting up a Jira Project to work with openedx-webhooks".  I'm also okay with something small in the README for now.

feanil (approved):
A couple of questions but generally looks good to deploy.

openedx_webhooks/settings.py
nedbat: I did this when I thought we'd have parallel environment variables for Jira settings, and it was awkward to parse them and mock them.  Now we aren't doing that, this does seem over-complicated.  I can undo it.

nedbat: Done

openedx_webhooks/tasks/pr_tracking.py
nedbat: As it turns out, we never use this value at the moment.  We'll get issues created, and then fix up the initial status somehow.  Ideally, we wouldn't have to specify the initial status, the project could just set it implicitly.

nedbat: I've deleted it.

feanil (approved):
:shipit:

PR #269, opened 2023-10-17
Run on 3.11, and upgrade one package at a time.

nedbat:


feanil (commented):
Looks like this re-compiles but does not do any upgrades yet?  Is the plan to upgrade individual things in this PR or in follow-ups?

nedbat:
I should have explained: I added the typing dependencies, then ran `make upgrade-package package=types-PyYAML` to only add the newly added packages.  So the generic upgrades haven't happened yet.  I would like to wait on the broad upgrades until we can review their changelogs.

feanil (approved):

PR #275, opened 2023-11-03
docs: how to configure labelogenic Jira issues

nedbat:
@feanil: I'm hoping these docs are enough that you (or someone else) could help (for example) eduNext get set up for their Jira instance.  Not that I won't get them set up, but I want to leave enough information.


nedbat:
cc @mariajgrimaldi

docs/install.rst
```diff
@@ -88,6 +68,56 @@ GitHub
...
+            # A textual description of the Jira instance, to use in comments.
+            # It will be used in a sentence like this:
+            # "I've created issue PROJ-123 in {description}."
+            description: the private MegaCorp Jira
+
+The mapping file is YAML that maps a repo name to two pieces of information: a
+Jira project and an issue type to create in that project. The file can be at
+any URL.  As shown above, a raw GitHub URL is convenient.  The file specifies
+repos, possibly with wildcards, and for each provides the project and issue
+type to use:
```
feanil, line 101:
```suggestion
The mapping file is YAML that maps a repo name to two pieces of information: a
Jira project and an issue type to create in that project. The file can be at
any URL.  As shown above, a raw GitHub URL is convenient.  The reason this is 
separate from other config is so that this mapping file can be managed by Megacorp
while the info file above with credentials in it is managed by the team running 
openedx-webhooks. The file specifies repos, possibly with wildcards, and for each
provides the project and issue type to use:
```

It might be useful to add a note about why there are two files here.

feanil (approved):
I think that's enough info for me to set this up for others.

PR #281, opened 2023-11-20
feat: add a `from-GitHub` Jira label to Jira issues. #279

nedbat:


docs/details.rst
```diff
@@ -128,3 +128,5 @@ bot will create a Jira issue in the Jira server with the "xyz" nickname.
 
 Each Jira server can specify a mapping of repos to other Jira details such as
 the Jira project for the issue, and the issue type to create.
+
+Jira issues created this way will have a "from GitHub" Jira label applied.
```
feanil, line 132: Does this label need to be created on the JIRA side before it can be applied? Or will it automatically get created when you try to apply it?

feanil (approved):
One question that might suggest a change in the docs but nothing else.

docs/details.rst
nedbat: It doesn't need to be made first, but testing that turned up a different problem, now fixed: no spaces allowed in Jira labels.

nedbat:
It works:
image

PR #340, opened 2025-01-24
fix: Don't mention the owner directly in the welcome message.

feanil:
We had backticks before and we removed them when we re-worded this.
However, we should put them back in.  Currently the maintainers get
mentioned and there fore start recieving notifications on all PRs
whether or not they are ready for their review.  This is causing a bit
of churn and waste as people have to check more PRs than necessary
especially since many PRs have known reviewers from the same team or via
FCs.

### DIFF
--- a/openedx_webhooks/templates/github_community_pr_comment.md.j2
+++ b/openedx_webhooks/templates/github_community_pr_comment.md.j2
@@ -1,7 +1,7 @@
 Thanks for the pull request, @{{ user }}!
 
 {% if owner %}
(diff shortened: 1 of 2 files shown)


kdmccormick (approved):
🙏🏻

xitij2000 (approved):
The backticks were removed because it seemed like pinging would be the right thing to do, but I think it makes sense to reintroduce them if it's causing issues.

itsjeyd (approved):
@feanil LGTM 👍

PR #440, opened 2026-07-24
build: modernize openedx-webhooks to use uv and pyproject.toml

farhan:
> [!IMPORTANT]
> PR implemented with the assistance of [Claude Code](https://claude.com/claude-code). Refined and validated before being submitted for code review.

Modernize \`openedx-webhooks\`
Part of https://github.com/openedx/public-engineering/issues/506

## Summary

- Replace \`setup.py\`/\`setup.cfg\` with \`pyproject.toml\` (PEP 621 static metadata)
- Switch from pip-compile to \`uv\` with PEP 735 dependency groups; commit \`uv.lock\`
- Retain pylint/isort/mypy as on master
- Coverage config moved into \`pyproject.toml\`
- Update CI to use \`astral-sh/setup-uv\`; SHA-pin all actions; split into separate \`run_tests\` (produces "Tests (ubuntu-latest, 3.12)" check) and \`quality_and_docs\` jobs

## Removed/Updated

**Deleted files:** \`setup.py\`, \`setup.cfg\`, \`requirements/\`, \`.coveragerc\`

**Removed Makefile targets:**

| Target | Reason |
|---|---|
| \`compile-requirements\` | Replaced by \`uv lock --upgrade\` in the \`upgrade\` target |
| \`pip-compile\` | Replaced by \`uv lock\` |
| \`pip-compile-upgrade\` | Replaced by \`uv lock --upgrade\` |

**Updated Makefile targets:**

| Target | Change |
|---|---|
| \`upgrade\` | Now runs \`edx_lint write_uv_constraints\` + \`uv lock --upgrade\` instead of \`pip-compile\` |
| \`upgrade-package\` | Now runs \`uv lock --upgrade-package \$(package)\` instead of \`pip-compile\` |
| \`install-dev-requirements\` | Now runs \`uv sync --group dev\` instead of \`pip-sync\` |
| \`pylint\` | Removed \`setup.py\` argument — it was the only root-level \`.py\` file and is now deleted |
| \`check-pyproject.toml\` | Renamed from \`check-setup.py\`; validates \`pyproject.toml\` via \`tomllib\` instead of running \`python setup.py check\` |

## Versioning
Static: \`version = "0.1.0"\` declared directly in \`pyproject.toml\` — this repo does not publish to PyPI, so \`setuptools-scm\` is not used and the version is bumped manually on each release tag.

## Important Notes
- This repo does not publish to PyPI, so \`python-semantic-release\` and \`release.yml\` were not added.
- \`src/\` layout was not adopted — this repo does not publish to PyPI, so the flat layout is retained.

## Dependency Upgrade Workflow
The existing \`upgrade-python-requirements.yml\` calls \`openedx/.github\`'s reusable workflow, which hardcodes \`ADD_PATHS="requirements"\` — but that only affects the fork-friendly path (Path B, \`peter-evans/create-pull-request\`). Since this repo has \`REQUIREMENTS_BOT_GITHUB_TOKEN\` configured, the workflow takes Path A (\`pull_request_creator\`), which discovers changed files from the git diff directly and ignores \`ADD_PATHS\` entirely. \`openedx/XBlock\` is already on this same uv-based setup and its weekly upgrade workflow runs successfully, picking up \`uv.lock\` changes without any issues — confirming this repo will behave identically after merge.

## Testing Notes
This PR has not been manually tested against the repo's own features. Testing relied on CI checks and local agent tooling (\`make install-dev-requirements\`, \`uv run tox -e py\`, \`uv run tox -e quality\`, \`uv run tox -e docs\`). Repo-owner is encouraged to run the repo's feature tests before merging.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Makefile
```diff
@@ -31,67 +31,23 @@ test-html-coverage-report: test ## Run tests and show coverage report in browser
...
-ifneq (, $(wildcard $(PRIVATE_OUT)))
-	pip-sync $(PRIVATE_OUT)
-else
-	pip-sync requirements/dev.txt
-endif
-	@# This run of mypy is to discover the missing type stubs, then we install them
-	-mypy $(TYPEABLE) > /dev/null
-	mypy --install-types --non-interactive
+	uv sync --group dev
+	uv tool install tox --with tox-uv
```
irfanuddinahmad: `uv tool install tox --with tox-uv` installs an unpinned, un-lockfiled global tox outside `uv.lock` — this is the same anti-pattern flagged elsewhere in this migration effort. Since `ci.yml` uses `uv sync --group ci` + `uv run tox` (the locked, pinned tox from the `ci` dependency-group) and no Makefile target here actually invokes bare `tox` either, this line looks unnecessary — recommend dropping it.

irfanuddinahmad:
Two more findings that aren't attachable as inline comments since the files aren't part of this diff:

**`Dockerfile` is now broken (not touched by this PR).** It still does `COPY requirements ./requirements` (line 36) — that directory is deleted by this PR — followed by `RUN make install-dev-requirements` (line 37). Even past the COPY failure, this Dockerfile never installs `uv` (only `pip install tox`), so `make install-dev-requirements` would fail with `uv: command not found` regardless. Nothing in CI builds this image, so it's currently invisible to the green checks, but it's genuinely broken for local dev via Docker. Needs either an update (install uv, drop the requirements COPY) or, if it's dead/unused, worth deleting instead.

**`.github/workflows/upgrade-python-requirements.yml`** calls `openedx/.github`'s reusable workflow, which hardcodes `ADD_PATHS="requirements"` for its fork-PR-creation step. Since `requirements/` is deleted here, the weekly scheduled dependency-upgrade job will keep "succeeding" while silently producing no real PRs (the `uv.lock`/`pyproject.toml` changes won't match that hardcoded glob). Can't be fixed from this repo alone, but worth a callout in the PR description as a known gap, and maybe a follow-up issue against `openedx/.github` to make that path configurable.

(Everything else checked out — changelog handling correctly N/A here since this repo uses `scriv`→`README.rst` and that's untouched, action pinning verified genuine and consistent with this repo's pre-existing convention, `__init__.py` version handling correct, dependency-groups correctly kept in `[project.dependencies]` rather than a `base` group — which I verified is actually required for Heroku's buildpack auto-deploy via `uv sync --no-default-groups` to keep working.)

farhan:
Thanks for the thorough review — all three points addressed in bd9a5ac.

1. **`uv tool install tox`** — dropped from `install-dev-requirements`. 
2. **Dockerfile** — fixed
3. **`upgrade-python-requirements.yml` / `openedx/.github` gap** — added a Known Gaps section to the PR description.

openedx_webhooks/__init__.py
```diff
@@ -12,7 +13,10 @@
 from werkzeug.middleware.proxy_fix import ProxyFix
 from werkzeug.utils import import_string
 
-__version__ = "0.1.0"
+try:
+    __version__ = version("openedx_webhooks")
+except PackageNotFoundError:
```
irfanuddinahmad: This `except` branch is unreachable in tests -- the package is always installed during test runs, so `version("openedx_webhooks")` never raises `PackageNotFoundError` here. That's exactly why `codecov/patch` and `codecov/project` are currently failing on this PR (60% patch coverage, project down 0.06% -- these 2 lines are the whole diff). Suggest adding `# pragma: no cover` on this line, same as the pattern already used cleanly elsewhere in this migration effort (those repos just didn't hit a coverage-threshold failure from it).

tox.ini
```diff
@@ -0,0 +1,29 @@
...
+allowlist_externals = make
+commands =
+    make test
+
+[testenv:quality]
+runner = uv-venv-lock-runner
+dependency_groups = quality
+allowlist_externals = make
+commands =
+    make pylint
```
irfanuddinahmad, line 20: `isort` is declared as a dependency in the `quality`/`dev` groups (pyproject.toml) but never actually invoked anywhere -- no Makefile target, and this `quality` env only runs `make pylint` + `make mypy`. Every sibling repo in this migration effort wires `isort --check-only` into its quality check; this one just installs it and never runs it. Worth either adding an `isort` step (e.g. a `make isort-check` target added to this `commands` block) or dropping the now-unused dependency.

irfanuddinahmad (commented):
Two findings from a fresh pass over the current diff (verified against the actual head commit and live CI checks, not just inspection).

irfanuddinahmad (approved):
All findings from both review passes addressed in bd9a5ac: dead `uv tool install tox` dropped, the uncovered `except` branch pragma'd, unused `isort` dep resolved, Dockerfile fixed to install uv and stop copying the deleted `requirements/` dir, and the `upgrade-python-requirements.yml`/`openedx/.github` `ADD_PATHS` gap called out as a known limitation. LGTM.

tox.ini
farhan: Fixed — dropped `isort` from both `quality` and `dev` groups in `pyproject.toml` and regenerated `uv.lock`. It had no parity with master (master never had isort at all) and was never invoked anywhere, so removing it was the right call.

.github/workflows/ci.yml
```diff
@@ -13,29 +13,52 @@ jobs:
     name: Tests
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: true
```
salman2013: I believe it's true by default. any specific reason to add this?

salman2013 (approved):

.github/workflows/ci.yml
farhan: removed, let's keep the default value

farhan:
@feanil PR is ready review/merge

.github/workflows/ci.yml
```diff
@@ -14,28 +14,49 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os:
-          - ubuntu-latest
-        python-version:
-          - "3.12"
+        os: [ubuntu-latest]
```
feanil: nit: This is an un-necessary change and makes it more annoying to add new versions to a list by just adding a new line, rather than editing an existing line.  The diff is easier to read when just adding a new line.

feanil (approved):
generally looks good, I made a small style fix and ran the make upgrade.

